PR #386, opened 2026-05-28
fix(providers): ANNEX B4 — no silent routing default/refill + accurate pricing

thesongzhu:
## What & why
Closes the real ANNEX P1/P2 provider findings on current `origin/main`:

- **B4-pricing (P1)** — env auto-detect default models silently fell to the generic `$1/$4` balanced fallback, so cost/budget labels were inaccurate. Added accurate, officially-sourced pricing for:
  - OpenAI `gpt-4o` (2.50/1.25/10.00), `gpt-4o-mini` (0.15/0.075/0.60)
  - DeepSeek `deepseek-v4-pro` (1.74/0.0145/3.48), `deepseek-v4-flash` (0.14/0.0028/0.28) + `deepseek-chat`/`deepseek-reasoner` aliases priced as V4 Flash
  - Google `gemini-2.0-flash` (0.10/0.025/0.40)
  - **Standard list prices** (not the DeepSeek promo expiring 2026-05-31) so cost/budget never silently under-reports later. Sources + dates recorded inline in the catalog.
- **B4-fallback-injection (P1)** — removed the startup block that silently refilled an empty `fallbackProviderIds` (overrode an explicit empty choice).
- **B4-auto-select (P2)** — removed boot-time silent default-provider selection by hardcoded `ROUTING_PRIORITY`. A **single** available provider is still auto-selected (single-key BYOK boots ready); **two or more** leave routing unconfigured so the user chooses explicitly (`resolveRoute` surfaces `PROVIDER_NO_ROUTING`). Candidates are the **union** of newly-detected + already-enabled providers (deduped) so a later single new env key can't silently become default among several configured providers.

`W-PROVIDER-ROUTING-001` is kept as a diagnostic-only warning (no silent action).

## Migration invariant
Users with a **saved default / fallback list are untouched** — both removed paths only ever fired when no default was persisted / fallbacks were empty. Only new multi-provider local setups change (now require explicit choice), exactly the locked decision.

## Documented limitation (non-hidden)
`groq` / `mistral` / `xai` / `openrouter` env-default models still resolve to the generic `$1/$4` fallback — those provider kinds have zero catalog entries. Out of this bundle's scope (goal names OpenAI + DeepSeek; Google was already first-class).

## Verification
- New: `friday-bootstrap-auto-routing.test.ts` (single→select, ≥2→null, union/second-boot regression, dedup, BYOK both sources).
- New pricing assertions prove gpt-4o/gpt-4o-mini/deepseek/gemini-2.0-flash no longer hit `$1/$4`; existing fallback test preserved.
- 14 focused tests pass; `tsc` clean on changed files; `git diff --check` clean; no secrets.
- Two independent read-only reviewers PASS (correctness + regression/no-overclaim/no-scope-creep).

refs: B4-fallback-injection, B4-auto-select, B4-pricing

### DIFF
--- a/src/hub/bootstrap/friday-bootstrap-auto-routing.ts
+++ b/src/hub/bootstrap/friday-bootstrap-auto-routing.ts
@@ -1,0 +1,66 @@
+import type { FridayProviderKind } from "../../providers/model/friday-provider.types.js";
+
+export interface FridayBootstrapRouteCandidate {
+  kind: FridayProviderKind;
+  id: string;
+}
+
+export interface FridayBootstrapAutoRoutingDecision {
+  defaultProviderId: string;
+  defaultModel: string;
+  fallbackProviderIds: string[];
+}
+
+/**
+ * Decide the boot-time default route from auto-detected / existing providers WITHOUT
+ * making a hidden choice on the user's behalf.
+ *
+ * Locked product decision: multiple keys/providers require explicit user choice; Friday
+ * must not silently pick a default provider or silently inject fallback providers.
+ *
+ * - Exactly one candidate: select it (single, unambiguous — preserves single-key BYOK).
+ *   No fallback providers are injected.
+ * - Two or more candidates: return `null`. Friday must NOT auto-pick a default by a hidden
+ *   priority order, nor fan out fallbacks. The user chooses explicitly; until then
+ *   `resolveRoute` surfaces `PROVIDER_NO_ROUTING`.
+ * - No candidates: `null`.
+ */
+export function resolveFridayBootstrapAutoRouting(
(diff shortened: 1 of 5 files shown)


thesongzhu:
Superseded by #384 (`cc3cefc6`, 'provider/settings truth + setup-safety closure'), which independently landed the same locked-decision provider-routing + pricing fixes on main while this was in review. Closing to avoid a redundant/conflicting change. The one net-new item here that #384 did not cover — accurate `gemini-2.0-flash` pricing (the GOOGLE env default, which #384 leaves at the new unknown/0 rate) — will be folded into the remaining provider-cost-truth follow-up on top of cc3cefc6.

PR #393, opened 2026-05-29
fix(agent,memory): destructive shell flags require approval + PII redaction of metadata/tags (ANNEX B6)

thesongzhu:
Closes two ANNEX B6 P2 tightenings not covered by the parallel orchestrator's merged B6 track (#388 completion / #390 evidence-durability / self-heal-stub removal).

## Dangerous shell flags require approval
Shell risk was classified by **program name** only, so `git reset --hard`, `git clean -fdx`, `git checkout --force`, and `find . -delete` slipped through as "safe" (no approval) because `git`/`find` are name-allowlisted.
- New `detectDangerousShellFlagReason(program, parts)` flags those destructive flag combinations.
- Wired into BOTH `classifyShellRisk` (→ `destructive`) and `getApprovalRequiredReasonForExecCommand` (→ approval gate).
- Benign git/find (`git status`, `git reset HEAD`, `git clean -n`, `find . -name …`) stay `safe` (no false-positives).

## PII redaction covers content, metadata, AND tags
Previously only `content` was scanned/redacted; caller-supplied metadata + tags (reachable via the HTTP store route) passed through raw.
- New `piiGuard.redactDeep(value)` recursively redacts PII in the string leaves of metadata objects and tag arrays (depth-bounded; non-strings untouched).
- **Store path**: redacts metadata + tags, merges discovered `pii.*` tags, and block-mode now fires on metadata/tag PII too.
- **Read path** (output-filter): redacts metadata + tags on read-back (defense in depth for pre-existing items).

## Verification
- Both paths proven: dangerous-flag → destructive/approval AND benign → safe/null; redactDeep redacts email/phone/SSN in nested metadata + tags, leaves clean values + non-strings untouched, tag-vs-redact mode.
- 414 memory/tool-risk/adversarial tests pass; `tsc` clean; `git diff --check` clean; secrets baseline line-number refresh (no new secrets).

refs: B6-shell-flags, B6-pii-metadata

### DIFF
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -603,7 +603,7 @@
         "filename": "test/unit/agent/runtime/friday-agent-tool-risk.test.ts",
         "hashed_secret": "d14babe099a92466c4efa49400806772a6c097cd",
         "is_verified": false,
-        "line_number": 162
+        "line_number": 185
       }
     ],
     "test/unit/api/auth/friday-auth-middleware.test.ts": [
@@ -1124,5 +1124,5 @@
       }
     ]
   },
-  "generated_at": "2026-05-28T23:15:45Z"
+  "generated_at": "2026-05-29T00:26:37Z"
 }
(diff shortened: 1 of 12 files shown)


thesongzhu:
Superseded by bundled ANNEX B6 PR #395 to avoid separate proof/merge cycles. #395 contains this shell-risk + memory PII work plus the related B6 connection-error UX work, after drift classification against current main.

PR #394, opened 2026-05-29
fix(ui): user-readable next steps for agent-run connection errors (ANNEX B6)

thesongzhu:
Closes the final ANNEX B6 item: "common setup/session/provider errors include user-readable next steps."

The agent-run SSE hook (`use-agent-run-events.ts`) surfaced raw technical strings — `Session expired`, `HTTP <status>`, raw exception message — with no guidance. Replaced with actionable copy:
- session expiry → "Your session expired — please sign in again to continue."
- non-OK response → "Couldn't reach Friday (HTTP <status>). Make sure the Friday server is running, then try again."
- transport error → "Lost connection to Friday — reconnecting automatically…" (hook already retries with backoff), preserving the underlying detail in parentheses.

## Scope / verification
- Display-copy only; no behavior/logic change (diff is 1 file, +6/−4).
- Verified by `tsc -p ui/tsconfig.json` (clean) and the CI UI `build` job. The repo has **no `ui/src` unit-test harness** (0 ui/src tests; the vitest `default` project only covers `test/**`), so there is no behavioral unit to add — the deliverable is the user-facing wording, which reviewers can verify directly.

refs: B6-error-ux

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #395, opened 2026-05-29
fix(annex-b6): bundle shell/PII safety and connection-error next steps

thesongzhu:
## Why
Bundle the ANNEX B6 residuals that were previously split across #393 and #394 so they share one drift classification, one proof cycle, one merge, and one exact-main closure.

## What
- Require approval/destructive classification for dangerous flags on otherwise allow-listed shell programs, including git global-option forms like `git -C repo reset --hard`, `git --no-pager clean -fdx`, and `find . -delete`.
- Extend memory PII handling to caller-supplied metadata and tags: redact metadata string leaves, drop PII-bearing tags, surface pii.* type tags, and apply read-path defense-in-depth filtering.
- Add user-readable next steps for agent-run SSE connection errors while preserving technical detail for reconnect/debug context.

## Drift classification
- #392 changed autonomous engine/types/tests only.
- #393 changes agent runtime shell-risk + memory guard/PII files.
- #394 changes `ui/src/hooks/use-agent-run-events.ts` only.
- No file overlap or proof-semantic conflict with #392; #393/#394 are same ANNEX B6 proof surface and are intentionally bundled here.

## Local proof
- `git diff --check origin/main...HEAD`
- `npm test -- --run test/unit/agent/runtime/friday-agent-tool-risk.test.ts test/unit/memory/guard/services/friday-memory-pii-guard.test.ts test/unit/memory/guard/services/friday-memory-guard-service-pii-policy.test.ts test/unit/memory/guard/services/friday-memory-guard-service-output-filter.test.ts test/unit/memory/guard/services/friday-memory-guard-service-validation.test.ts` — 5 files, 124 tests, Type Errors none
- `npx tsc -p ui/tsconfig.json --noEmit`
- `npm run --silent build:api`
- `npm run --silent check:secret-patterns`

## Boundary
No npm publish, tag, GitHub release, live-channel window, GitHub secrets/settings change, destructive user-data mutation, force/admin bypass, or final stop-boundary decision.

### DIFF
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -603,7 +603,7 @@
         "filename": "test/unit/agent/runtime/friday-agent-tool-risk.test.ts",
         "hashed_secret": "d14babe099a92466c4efa49400806772a6c097cd",
         "is_verified": false,
-        "line_number": 162
+        "line_number": 185
       }
     ],
     "test/unit/api/auth/friday-auth-middleware.test.ts": [
@@ -1124,5 +1124,5 @@
       }
     ]
   },
-  "generated_at": "2026-05-28T23:15:45Z"
+  "generated_at": "2026-05-29T00:26:37Z"
 }
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -603,7 +603,7 @@
         "filename": "test/unit/agent/runtime/friday-agent-tool-risk.test.ts",
         "hashed_secret": "d14babe099a92466c4efa49400806772a6c097cd",
         "is_verified": false,
-        "line_number": 185
+        "line_number": 201
       }
     ],
     "test/unit/api/auth/friday-auth-middleware.test.ts": [
@@ -1124,5 +1124,5 @@
       }
     ]
   },
-  "generated_at": "2026-05-29T00:26:37Z"
+  "generated_at": "2026-05-29T00:48:14Z"
 }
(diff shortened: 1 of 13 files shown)


thesongzhu:
**Adversarial-review finding carried over from the now-superseded #393 (shell-risk portion of this bundle).** Two reviewer rounds on #393 closed the original `git -C <path> reset --hard` bypass, but a second-pass review found a **residual no-default-allow gap** that this bundle inherits verbatim (`src/agent/runtime/friday-agent-tool-risk.ts`):

`gitSubcommandStart()` allow-lists known git global options in `GIT_GLOBAL_OPTS_WITH_VALUE` and treats any **unlisted** leading option as single-token (skip 1). Git global options that take a **separate-token value** but are absent from the set therefore shift the subcommand and bypass the gate. Confirmed against git 2.39.5 + the classifier:
- `git --shallow-file p reset --hard` → classifies **safe**, `getApprovalRequiredReasonForExecCommand` → **null** (no approval)
- `git --config-env k=v reset --hard` → same bypass

`--shallow-file <path>` / `--config-env <name>=<envvar>` / `--attr-source <tree>` are documented top-level git options. This is a default-allow fallback on unknown options (locked decision: no default-allow). Reachability in agentic output is low (the common forms `-C`/`-c`/`--no-pager`/`--git-dir=`/`--work-tree` ARE covered), so it's MAJOR-not-blocker, but it is the same mechanism the round-1 blocker was about.

**Recommended fail-safe fix** (no-default-allow): don't allow-list — instead treat an unrecognized space-form leading `--option` as value-consuming, OR scan the full post-`git` token list for the destructive `(subcommand, flag)` combination regardless of preceding global options. Add probe tests: `git --shallow-file x reset --hard` and `git --config-env k=v reset --hard` must be destructive + approval-required, with `git --shallow-file x status` staying safe.

(Minor, also from the review: stale comment in `friday-memory-guard-service.ts` says PII_MODE default is "tag" — it's "redact". Pre-existing, non-blocking.)

PR #396, opened 2026-05-29
fix(agent): fail-safe git global-option parsing in shell-risk gate (ANNEX B6 follow-up)

thesongzhu:
## What
Closes the residual default-allow bypass in the dangerous-shell-flag approval gate that shipped with the now-merged ANNEX B6 bundle (#395). Reviewer A flagged it on #395; the orchestrator merged #395 as-is, so this is a clean follow-up off current `main`.

## The gap
`gitSubcommandStart` located the git subcommand by skipping a fixed allow-list of global options and treating any **unlisted** leading option as single-token. Git plumbing options that take a **separate-token value** therefore shifted their value into the subcommand slot:

```
git --shallow-file snap reset --hard   # classified SAFE, no approval (BYPASS)
git --config-env  k=v   reset --hard   # same
git --attr-source tree  reset --hard   # same
```

Common forms (`-C`, `-c`, `--no-pager`, `--git-dir=`, `--work-tree`) were already covered, so reachability was low — but it was a **default-allow fallback on unknown options**, which the ANNEX repair goal explicitly forbids ("do not close with: default-allow fallback").

## The fix
`gitSubcommandStart` now fails safe:
- recognized value-options skip their value (`-C <v>` → +2);
- recognized value-less flags skip one token (`--no-pager` → +1);
- `--opt=value` is one token;
- **any UNRECOGNIZED leading option** drops to a full-token scan that detects a destructive `(subcommand, flag)` combination regardless of the options preceding it.

The named option sets are now an **optimization for the precise path, not a security boundary** — a forgotten/future option degrades to the scan, never to a bypass. This converts Reviewer A's "default-allow on unknown options" into fail-safe behavior.

No false-positive: the scan runs **only** behind an unrecognized leading option, so `git commit -m "reset --hard"` stays safe (the `commit` subcommand is found by position; the scan is never reached).

## Verification
- Focused suite `friday-agent-tool-risk.test.ts`: **53 pass** (incl. new probes for `--shallow-file`/`--config-env`/`--attr-source`, a genuinely-unknown option, no-over-block, and the commit-message false-positive guard).
- Blast radius `test/unit/agent/{runtime,tools}`: **1263 pass**, no type errors.
- `git diff --check` clean; eslint 0 errors; detect-secrets no new findings (baseline line-number refresh only).

Proves both paths: unsafe (obscure + unknown options behind `reset --hard`/`clean -fdx`) requires approval; safe (benign subcommands behind the same options, `reset HEAD file`, `clean -n`, commit messages) stays safe.

### DIFF
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -603,7 +603,7 @@
         "filename": "test/unit/agent/runtime/friday-agent-tool-risk.test.ts",
         "hashed_secret": "d14babe099a92466c4efa49400806772a6c097cd",
         "is_verified": false,
-        "line_number": 201
+        "line_number": 227
       }
     ],
     "test/unit/api/auth/friday-auth-middleware.test.ts": [
@@ -1124,5 +1124,5 @@
       }
     ]
   },
-  "generated_at": "2026-05-29T00:48:14Z"
+  "generated_at": "2026-05-29T01:42:24Z"
 }
--- a/src/agent/runtime/friday-agent-tool-risk.ts
+++ b/src/agent/runtime/friday-agent-tool-risk.ts
@@ -141,22 +141,82 @@ function listPotentialFilePaths(text: string): string[] {
   return [...new Set(text.match(/\b[\w./-]+\.[A-Za-z0-9]+\b/g) ?? [])];
 }
 
-// git global options that precede the subcommand; the value-taking ones (given without `=`)
-// also consume the following token. Used to locate the real subcommand for forms like
-// `git -C /repo reset --hard` or `git --no-pager clean -fdx`.
+// git global options that precede the subcommand and, when given in their space-form
+// (without `=`), ALSO consume the following token. Used to locate the real subcommand for
+// forms like `git -C /repo reset --hard`. NOTE: this set is an optimization for the precise
+// path, not a security boundary — an option we forget to list does NOT open a bypass, because
+// `gitSubcommandStart` fails safe on anything it does not recognize (see below).
 const GIT_GLOBAL_OPTS_WITH_VALUE = new Set([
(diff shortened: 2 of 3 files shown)


thesongzhu:
## Review summary (two read-only reviewers)

**Reviewer B (regression/correctness): PASS** — no regression, behavior-preserving on the precise path, refactor is a verbatim lift of the prior inline logic, both call sites integrate correctly, 53 focused + 1263 blast-radius tests pass, `.secrets.baseline` change is line-number/timestamp only.

**Reviewer A (adversarial): the CHANGE PASSES** — verified empirically (transpiled the classifier, cross-checked every form against real git 2.39.5): "the PR's actual change … is sound and well-tested", "No bypass in the changed parse logic". Confirmed `--shallow-file`/`--config-env`/`--attr-source` + a genuinely-unknown option all require approval via the fail-safe scan; named sets are an optimization not the boundary; tests are real (not mocks) and exercise the unknown-option path; no new false-positive (commit-message guard holds).

Reviewer A returned an overall FAIL on **one finding that is out of this PR's scope**: path-qualified programs (`/usr/bin/git reset --hard`, `/bin/rm -rf`) bypass program *identification* (`parts[0]` is not basename-normalized) — which defeats the whole gate including the pre-existing `rm`/DESTRUCTIVE_PROGRAMS check. This is **pre-existing on `origin/main`** (the `program === "git"` literal predates this diff; the same gap was tagged "PRE-EXISTING and out of B6 scope" in the #395/#393 reviews). This PR does not touch or regress program identification. It is a real, separate no-default-allow gap that is being tracked separately for a decision — **not** introduced or worsened here.

This PR is scoped to the git global-option fail-safe parse, which both reviewers confirm is correct, regression-free, and bypass-free.

PR #398, opened 2026-05-29
audit C: run-level workflow completion-verification enforcement + durable persistence (Stage 1 + 2B)

thesongzhu:
## What & why

Audit finding **C** (workflow-node false-completion): a side-effect / mutating / user-visible-claim workflow node could be read as a **verified** completion off nothing more than arbitrary non-empty output. This PR implements **actual run-level enforcement** (not mere labeling) that closes the 3 run-level bypasses:

1. arbitrary non-empty output passing the vacuous acceptance baseline,
2. `output == null` skipping the gate entirely,
3. the disabled-pipeline / legacy path being treated as verified.

## Design: a SEPARATE, orthogonal truth axis (no reverse lie)

Completion-verification is tracked as its **own** run-level axis, fully **orthogonal** to evidence-persistence health (`evidenceStatus`). An earlier attempt overloaded `evidenceStatus → "degraded"` for any non-verified node — that was a *reverse lie* (a run with perfectly healthy persistence reported as "degraded") and broke 4 e2e proof/persistence tests. That overload is reverted; the two axes never conflate.

- **Runtime** (`friday-workflow-runtime`): per-run `completionVerification` aggregate (worst node label wins); `getRunCompletionVerification` is **fail-closed for non-settled runs** (a run not terminally `completed` is never `verified` → closes the mid-flight time-of-check race at the source). Surfaced on the evidence facade and the run receipt; never touches `evidenceStatus`.
- **Classification** (Stage 1, from part-1): a node is `verified` only when it is deterministic pure-compute or read-only; `ai`/`approval` → `model_assessed_unverified`; a side-effect node lacking deterministic evidence (or with unresolvable/empty grants — fail-closed) → `proof_pending`. Derived from declared capability (node type + manifest grants), **never** from output.
- **Task-workflow verifier** (the consumer / teeth) — **FAIL-CLOSED**: `verifyClaim` refuses a `workflow_run_evidence` ref whose source run is not `verified`, with the **distinct** code `TASK_WORKFLOW_CLAIM_WORKFLOW_RUN_COMPLETION_UNVERIFIED` (409) — never the persistence code. A *missing* completion lookup is itself a refusal, so losing the hub wiring becomes a **loud refuse** in production, never silent non-enforcement. The hub bootstrap wires it.

No closeout gate is added: `attachEvidenceRef` already blocks attaching refs to verified claims, and the completion label is immutable within a process lifetime, so (unlike persistence durability) there is no post-verify drift to re-check.

## Proof

- **Full default suite green:** 12226 passed / 159 skipped / **0 failed**; typecheck clean.
- **NEW enforcement e2e** (`...completion-verification-enforcement.acceptance`): side-effect & unresolved-skill runs are refused at `verifyClaim` with the distinct code; read-only run verifies (no over-blocking); **mid-flight TOCTOU** refused.
- **NEW service unit guards:** fail-closed when the completion lookup is absent (catches hub-wiring loss); `proof_pending` run refused distinctly from persistence.
- **run-enforcement unit:** asserts orthogonality (`proof_pending` AND `evidenceStatus === "available"`) across all 3 bypasses.
- The 4 prior blast-radius e2e tests pass; genuinely-informational fixtures were given read-only manifests + the completion lookup wired exactly as the hub wires it.

## Known limitation (scoped to Stage 2)

The completion aggregate is **process-lifetime in-memory** (same as `evidenceStatus`). After a hub restart, a previously `proof_pending` *completed* run reads `verified` (unknown→verified default). This **matches `evidenceStatus`'s existing behavior and is not a regression**; persisting the label durably is **Stage 2** (which is also where deterministic side-effect evidence would upgrade `proof_pending → verified`).

## Drift

Rebased onto latest `origin/main` (includes #395). No file overlap with concurrent work.

### DIFF
--- a/context/MEMORY.md
+++ b/context/MEMORY.md
@@ -18,4 +18,4 @@
 - Tool call summary (`src/agent/services/friday-tool-call-summary.ts`) captures privacy-safe tool execution metadata (arg keys only, no values) for observability and world model training data.
 - Warn-once pattern is used across 7+ modules (hub-bootstrap, auth, memory, system, http-server, workspace-context, unix-socket-bridge) to deduplicate runtime warnings without losing critical signals.
 - OpenAI Responses API (`openai-responses`) streaming is now supported alongside `openai-completions` in the agent LLM client.
-- Database migration count: 88 (latest: v088-auto-fix-rollback-receipt).
+- Database migration count: 89 (latest: v089-workflow-completion-verification).
--- a/src/hub/friday-hub-bootstrap.ts
+++ b/src/hub/friday-hub-bootstrap.ts
@@ -3531,6 +3531,14 @@ export async function createFridayHub(
       // repository or pretending unknown runs were healthy.
       getWorkflowRunEvidenceStatus: (runId) =>
         workflowRuntime.evidence.getRunEvidenceStatus(runId),
+      // Audit C: bridge the runtime's ORTHOGONAL run-level
+      // completion-verification truth (a side-effect node lacking
+      // deterministic evidence → `proof_pending`). verifyClaim refuses a proof
+      // claim backed by a non-verified run for a reason DISTINCT from
+      // persistence durability. Wiring it here makes the run-level enforcement
+      // live in production (the verifier is fail-closed once wired).
+      getWorkflowRunCompletionVerification: (runId) =>
+        workflowRuntime.evidence.getRunCompletionVerification(runId),
     });
     return { service, disabledReason: null };
   })();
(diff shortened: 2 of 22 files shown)


thesongzhu:
## Update: Stage 2 resolved — this PR is C **enforcement + persistence** (Stage 1 + 2B)

Rebased/merged onto latest `main` (includes #396/#397/#399 shell-risk + **#401 D1 receipt-truth**, all adopted as `closed_by_remote_change`).

**Stage 2(B) added (durable label persistence):** migration `v089` adds nullable `workflow_runs.completion_verification`; the runtime persists the worst non-verified label and reads in-memory ⊔ persisted (worst wins), so a `proof_pending` run stays `proof_pending` across a hub restart (amnesia fixed). New `...completion-verification-persistence.test.ts` proves it with a restart simulation.

**Stage 2(A) evidence→verified is deliberately NOT here.** It was verified out as **forgeable** without a runtime-observed receipt: workflow nodes carry no runtime-trustable side-effect evidence (skill returns only `{output}` → `data:` URI; no on-disk state; no config-declared target), so any re-read/re-checksum re-hashes the skill's own bytes. Honest evidence→verified requires the skill-adapter/`invokeSkill` contract change (touches every action-node path) → **separate scoped PR with a blast-radius estimate first**, not indefinitely deferred.

**C status after this merges:** ENFORCEMENT-complete / safety-property-closed (no false-verified; unverified side-effect runs stay `proof_pending` and are refused as proof backing; label durable). The verified-**upgrade** path remains `proof_pending` for all side-effect nodes, scheduled as contract-change (a).

Local: full suite GREEN (12238 passed/0 failed), lint 0 errors, dist rebuilt, both RGG public-surface scenarios pass.

PR #404, opened 2026-05-29
fix(self-heal): durably persist self-heal skill-status so disabled skills survive restart

thesongzhu:
## Summary

Residual E (self-repair safety/recovery semantics). Self-heal `disable_skill` / `regenerate_skill` wrote skill lifecycle status **only** to the process-local `createStubMemoryState` stub, so a skill that self-heal **disabled** (because it was failing/unsafe) **silently re-enabled on the next hub restart** — and a regenerated/re-enabled skill's status could likewise be lost.

This wires **both** self-heal executors to the **existing durable** skill-lifecycle store so the decision survives a restart.

## Change (scoped per approved plan)

- `createFridayHubAutoFixExecutionSupport` gains two **optional** deps: `persistSkillLifecycleStatus(id, status) -> boolean` and `getPersistedSkillLifecycleStatus(id)`.
- `disable_skill` and `regenerate_skill` executors call the durable persister after the in-memory write; the verifiers use a durable-aware read (`resolveSkillLifecycleStatus`).
- The hub bootstrap binds the persister to `converterSkillRepo` — `getSkillById` existence check, then `updateLifecycleStatus` inside `withWriteTransaction`.

**Deliberately narrow:**
- Persists **only** for skills already in the durable `skills` table (installed/converter skills). Built-in / registry-only skills are **not upserted** — `persistSkillLifecycleStatus` returns `false` (no fake write), and the verifier reports an **honest non-durable result** (falls back to in-memory) instead of false-green.
- **No migration, no schema change, no registry-population change.**

## Proof

`test/unit/hub/bootstrap/hub-helpers.test.ts` (16 pass, types clean):
- **disable → restart** (fresh empty `memoryState`, same durable store) → still `disabled`.
- **regenerate/re-enable → restart** → `installed` (no stale `disabled` — catches the counterpart-staleness trap).
- **negative**: a built-in / not-in-store target is **not** fake-persisted (`store` has no entry, `_skillStatusDurable=false`), and the verifier honestly returns `false` after restart.

No-regression: existing self-heal/auto-fix suites pass unchanged (durable hooks are optional) — `friday-auto-fix-execution-service`, `friday-auto-fix-rollback-service`, `friday-self-healing-api-service` (34 tests). `build:api`, `check:secret-patterns`, `git diff --check` clean.

## Note

For in-store skills this also makes a self-heal disable effective for the durable-read workflow/install gates **in-process** (not only after restart) — the intended safety behavior (a self-heal-disabled skill is blocked, not silently runnable).

## Scope boundary

Standalone — does not bundle the other E truth-label items (FRIDAY_NODES_ENABLED comment, Desktop-Recording UI copy). No npm publish, tag, release, secrets/settings, or migration.

### DIFF
--- a/src/hub/bootstrap/hub-helpers.ts
+++ b/src/hub/bootstrap/hub-helpers.ts
@@ -1177,12 +1177,36 @@ export function createFridayHubAutoFixExecutionSupport(deps: {
   workflowRuntime?: FridayWorkflowRuntime;
   skillGenerator?: FridaySkillGeneratorService;
   nowIso: () => string;
+  // Durable skill-lifecycle persistence (optional). When provided, self-heal status
+  // mutations (disable_skill / regenerate_skill) are persisted to the durable skills store
+  // so a disabled or re-enabled skill survives a hub restart instead of living only in the
+  // process-local memoryState stub. `persistSkillLifecycleStatus` returns true ONLY when the
+  // skill already exists in the durable store (installed/converter skills); for built-in /
+  // registry-only skills it returns false (no fake write), and the in-memory status remains
+  // the only record. `getPersistedSkillLifecycleStatus` returns undefined for a not-in-store
+  // skill so verifiers can report an honest (non-durable) result.
+  persistSkillLifecycleStatus?: (skillId: string, status: SkillLifecycleStatus) => boolean;
+  getPersistedSkillLifecycleStatus?: (skillId: string) => SkillLifecycleStatus | undefined;
 }): FridayHubAutoFixExecutionSupport {
   // P2-07: External-state remediation must be backed by hub-level services.
   // The execution service fails closed for these kinds unless an executor is
   // injected here.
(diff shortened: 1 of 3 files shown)


thesongzhu:
Superseded by #406 (merged as \`dbaf77b403\`), which landed the equivalent narrow self-heal skill-status durability fix: explicit lifecycle transitions persist to the \`skills\` table via \`updateLifecycleStatus\` (the source the workflow-execution safety gate reads), while registry discovery (\`upsertDiscoveredSkills\`) stays in-memory so it can't clobber the converter's \`not_installed\`. Same goal + same safe approach as this PR. Recommend closing as superseded to avoid a double-implementation conflict on \`main\`. (Flagging per concurrency coordination; leaving the close to its author/maintainer.)

thesongzhu:
Superseded by the capture-and-restore rework. The original blind durable-write had a validated clobber risk (writing 'installed' could promote an unpromoted 'not_installed' candidate, defeating the workflow safety block — as #405 flagged). Reopening as a plan-build-time capture-restore implementation per the agreed design.

PR #421, opened 2026-05-30
fix(workflows): parse-validate data-node expressions in the generated-workflow validator

thesongzhu:
## Problem (root cause)
Workflow generation-to-run was nondeterministic (~37.5% generate-run pass across 16 attempts, independent of hub sharing). Captured failing-run node errors show the exact blocker:

- `EXPRESSION_PARSE_ERROR: unexpected identifier 'sum'; use $sum for references`
- `EXPRESSION_PARSE_ERROR: unexpected character '+' at position 13`

The generator's pre-approval validator (`friday-generated-workflow-validator.ts`) checked that data nodes *have* `config.mapping`/`config.transform` (presence) but **not** that the expression is **parseable** by the run engine. The node executor `exec()`s `config.transform` and `$`-prefixed `config.mapping` values (`resolveArgs`). DeepSeek sometimes emits invalid expressions (`sum(...)` vs `$sum(...)`, malformed operators) — structurally valid, so they pass approval, then throw `EXPRESSION_PARSE_ERROR` at run under fail_fast.

## Fix
Parse-check `config.transform` and `$`-prefixed `config.mapping` values in the validator, reusing the engine's pure-syntax `parse()` (tokenize+parse, no evaluation). Invalid expressions now fail validation → the existing generator repair loop relays the parse error → the model can fix it, instead of approving a workflow that explodes at run.

- **Fail-closed hardening — strengthens validation, never weakens it.** No schema is hardcoded.
- `parse()` is pure syntax: valid-but-unresolved refs (`$steps.x`) still pass; only true syntax errors fail. **No false positives** on literal mapping strings (not `$`-prefixed → not evaluated → not checked).
- Scoped to the reproduced data-node `transform`/`$`-mapping failure mode. `condition` (`conditionExpr`) and `ai`-node (`$ref` interpolation) expressions are noted follow-ups.

## Proof
- **Unit (deterministic) fail-before/pass-after** (`friday-generated-workflow-validator.test.ts`, +4 tests): invalid `transform` `sum(1,2)` and invalid `$`-mapping `$a + + b` now produce `GRAPH_DATA_NODE_EXPRESSION_PARSE_ERROR`; valid `$steps.a.value` + literal `"hello world"` do NOT. 16/16 pass.
- **Live (DeepSeek deep-proof) fail-before/pass-after** over 10 NL prompts: generate-run reliability **~37.5% → ~88%** (14/16 across 2 post-fix runs; safety behaviors 100%). A residual ~12% non-parse runtime failure mode remains (separate follow-up).
- **Blast radius:** 442 workflow tests pass (0 fail); validator unit 16/16; typecheck exit 0; lint 0 errors.

### DIFF
--- a/src/workflows/generator/validation/friday-generated-workflow-validator.ts
+++ b/src/workflows/generator/validation/friday-generated-workflow-validator.ts
@@ -12,6 +12,7 @@ import {
   getFridayWorkflowStepIdFormatMessage,
   isFridayWorkflowStepIdExpressionSafe,
 } from "../../utils/friday-workflow-step-id.js";
+import { createFridayExpressionEvaluator } from "../../engine/friday-workflow-expression-evaluator.js";
 
 // ─── Interface ───
 
@@ -315,6 +316,27 @@ export function createFridayGeneratedWorkflowValidator(
       // ─── Graph validation (if compiled) ───
 
       if (compiledGraph) {
+        // Pure-syntax parser shared across data nodes. parse() tokenizes +
+        // parses WITHOUT evaluating, so valid-but-unresolved refs (e.g.
+        // $steps.x) pass — only true syntax errors throw. This catches the
+        // runtime EXPRESSION_PARSE_ERROR class (e.g. `sum(...)` instead of
+        // `$sum(...)`) at validation time so the generator repair loop can
+        // relay it, instead of approving a workflow that explodes at run.
+        const expressionEvaluator = createFridayExpressionEvaluator();
+        const parseCheck = (expr: string, nodeId: string, where: string): void => {
+          try {
+            expressionEvaluator.parse(expr);
+          } catch (err) {
+            const message = err instanceof Error ? err.message : String(err);
+            issues.push({
+              code: "GRAPH_DATA_NODE_EXPRESSION_PARSE_ERROR",
+              stage: "graph",
+              severity: "error",
+              message: `Data node "${nodeId}" ${where} has an invalid expression: ${message}`,
+              stepId: nodeId,
+            });
+          }
+        };
         for (const node of compiledGraph.graph.nodes) {
           if (node.type !== "data") {
             continue;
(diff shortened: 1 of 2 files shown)


thesongzhu:
**Deferring this PR — root-cause captured, but a correct fix needs a product-direction decision.**

Empirical follow-up surfaced that the parse-check has a false positive and the underlying issue is deeper than this validator:

- The data-node executor runs `transform` via `expressionEvaluator.exec()` = `parse()` + evaluate, with **no JSON fast-path**. The expression evaluator is **reference-resolution-only**: `$a.b.c` and `'literal'` parse, but `{"k":"v"}` (object literal), `$a + $b` (arithmetic), and `$fn($a,$b)` (function calls) all **`EXPRESSION_PARSE_ERROR`**.
- So a `{"message":"version two"}` transform (which `test/unit/workflows/generator/services/friday-workflow-generator-service.test.ts` documents the generator *normalizing into existence*) actually **fails at run** — yet this parse-check flags it at validation, breaking that test (CI `test`/`quality-gate` red).
- That means the generator emits transform forms the evaluator can't execute. The correct fix is a **product-direction call** — either make the evaluator support object-literal/arithmetic transforms, or constrain the generator/normalizer to evaluator-compatible transforms — not a validator band-aid.

**Root cause (precise) stands and is the deliverable:** workflow-gen→run failures are `EXPRESSION_PARSE_ERROR` from a presence-vs-parseability gap (validator checked that data nodes *have* mapping/transform, not that the expression *parses*) compounded by a reference-only evaluator. Workflow-generation-to-run remains **proof_pending / non-claimed** for 1.0.3 (matches the established verdict). Recommend the product-direction fix as a follow-up before claiming workflow-gen reliability.

thesongzhu:
Superseded by #423 (merged as cf0a0db1), per operator decision. #423 includes this PR's data-node expression parse-check AND fixes its false positive (a JSON-object string transform is now normalized to a runnable `mapping` instead of being flagged), plus the real evaluator/executor fix (arithmetic + nested-mapping recursion) and adversarial-security hardening. Closing as superseded.

PR #424, opened 2026-05-30
fix(workflows): run object-literal STRING data transforms to completion

thesongzhu:
## Root cause (workflow-generation-to-run)
The generator legitimately emits object-valued data transforms as JSON-object **string** literals (the closure hello-world used `transform: '{"text":"hello world"}'`). The data-node executor passed such a string straight to the expression evaluator, which only handles scalar/ref/comparison expressions and throws `EXPRESSION_PARSE_ERROR: unexpected character '{'`. So the workflow passed validation (presence check) but **exploded at run** — the precise root cause behind `local.workflows.generator` / `local.uix.deploy-export` NO-GO.

## Fix (runtime correctness, fail-safe, zero-regression)
In the data node: if a `transform` string parses as a plain JSON object, resolve it through the existing, already-tested **mapping** path (`resolveArgs`: `$`-refs evaluated, literals passed through). Non-object transforms (refs, quoted literals, scalars, arrays) fall through to expression evaluation exactly as before. Every object-literal string previously **threw**, so no previously-working input changes behavior.

## Deliberately NOT changed (product-direction)
Arithmetic (`$a + $b`) and function-call (`sum(...)`) transforms still **fail-closed** at run. That is a separate expression-language decision (richer evaluator vs constrained generator), documented as a follow-up proposal — **not shipped in 1.0.3**. No validator weakened; no teach-to-test.

This fixes the deterministic **runtime-correctness** half. Generation **reliability** (whether the live model emits a runnable workflow on a given run) is nondeterministic and out of scope here.

## Tests (deterministic, output-VALUE oracles — not run-to-completion-only)
- `friday-workflow-node-executor.test.ts`: object-literal string transform (constant + `$`-ref resolution, proving real resolution not constants) + preserved expression/quoted-literal behavior. Fail-before reproduced the exact `EXPRESSION_PARSE_ERROR`; pass-after green.
- `friday-generated-workflow-transform-runtime.test.ts`: frozen corpus through the **real compiler + executor**, incl. arithmetic/function BOUNDARY cases that must still fail-closed.
- Blast radius: 751 workflow unit tests pass; typecheck clean.

### DIFF
--- a/src/workflows/engine/friday-workflow-node-executor.ts
+++ b/src/workflows/engine/friday-workflow-node-executor.ts
@@ -86,6 +86,38 @@ function resolveArgs(
   return resolved;
 }
 
+/**
+ * The workflow generator legitimately emits object-valued data transforms as
+ * JSON-object *string* literals (e.g. `'{"text":"hello world"}'`,
+ * `'{"greeting":"$inputs.name"}'`). The expression evaluator only handles
+ * scalar/ref/comparison expressions and throws `EXPRESSION_PARSE_ERROR` on an
+ * object literal, so such transforms used to fail at run despite passing
+ * validation. If a transform string parses as a plain JSON object, return it so
+ * the caller can resolve it via the same already-tested mapping path ($-refs
+ * evaluated, literals passed through). Returns undefined for anything that is
+ * not an object literal (refs, quoted literals, scalars, arrays) so those fall
+ * through to expression evaluation exactly as before — zero behavior change for
+ * inputs that already worked (all object-literal strings previously threw).
+ */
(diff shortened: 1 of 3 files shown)


thesongzhu:
Superseded by #423 (merged as cf0a0db1), per operator decision. #423 is the canonical workflow-gen→run fix: it covers the same object-literal-string transform runtime-correctness as this PR AND adds arithmetic + nested mapping + validator parse-check (faithful oracle on the richer grammar, supersedes #421's false positive). This PR's real-green-gate was UNSTABLE; #423 is fully green. Closing as closed_by_remote_change.

PR #589, opened 2026-06-08
feat(rust-fs): read-only `search` fs tool (safe-fs READ parity)

thesongzhu:
## What

Wires the registered-but-unimplemented `search` agent-loop fs tool to a new **READ-ONLY** `friday_fs::search_within_root` backend — a literal-substring grep contained to the workspace root. Completes safe-fs **READ** parity (read_file / list_dir / stat_file / **search**).

## friday-fs (`crates/friday-fs/src/lib.rs`)
- `search_within_root(root, query, Option<subpath>) -> Result<Vec<SearchHit>, FsError>`: recursive literal-substring (`str::contains`, case-sensitive) search over UTF-8 text files.
- **Containment (no new attack surface):** every on-disk access reuses the EXISTING hardened primitives — `list_dir_within_root` (O_NOFOLLOW + realpath-ancestor), `stat_file_within_root` (no-follow lstat), `open_read_within_root` (O_NOFOLLOW + fd-identity TOCTOU). Candidate paths are built only from `readdir` entry names under the contained root, so each access is independently re-validated by `resolve_within_root`. **No change to the containment pipeline or the gate.**
- **Symlinks never followed:** every entry is stat'd before it is touched; a symlink → `Err(Symlink)` → SKIPPED, so an outside-pointing link is never read or descended. An absolute / `..` / out-of-root / final-symlink `subpath` argument is rejected (`Err`). Binary/non-UTF-8 files skipped (NUL screen + valid-UTF-8-prefix decode).
- **Bounded output (documented consts):** `SEARCH_MAX_HITS=200`, `SEARCH_MAX_LINE_BYTES=512` (char-boundary truncation), `SEARCH_MAX_FILES=2000`, `SEARCH_MAX_DIRS=4096`, `SEARCH_MAX_FILE_BYTES=1 MiB` via `Read::take` (holds even if a file grows after stat — never unbounded memory/output).
- `tests/search_fs.rs`: 11 real-fs adverse/bounds tests (recursive happy path; scoped subpath; single-file subpath; absolute/`..`/ancestor-symlink/final-symlink subpath rejection; **real symlink-entry-not-followed, outside content not leaked**; hit cap; per-line char-boundary truncation; binary skip; empty query).

## friday-hub (`crates/friday-hub/src/lib.rs`)
- `"search"` arm in `FsToolExecutor::execute` → **read-type receipt**: matches in `content` (`relpath:line:text`, model-facing); `summary` is a **COUNT only** so matched lines never enter the hash-chained audit ledger (mirrors list_dir/read_file). Optional `path` param scopes the search. `search` was already registered `mutating=false` / `Risk::ReadOnly` → **gate Allows, no approval path**.
- The two tests that used `search` as the gate-Allowed-but-Unsupported example now trigger the exec-error path via `read_file`-on-missing (search is wired). New tests: search executes through the full turn loop (`Allow` + `executed`), direct search receipt (content matches + count-only summary), and `run_command` still `Unsupported` (preserves the fail-closed `other =>` arm coverage).

## Truth label
Read-only `search` completes safe-fs **READ** parity; gate-Allowed (non-mutating); honors the full workspace containment pipeline; bounded output. **PROOF-ONLY — not v1 GO.** `executeRun` is not replaced; no gate / schema / migration / mutating change; **no TS**; `run_command` remains a separate later slice. Known test-coverage boundary: the files-scanned / dirs / per-file-byte caps are implemented + documented but not externally observable through the public API, so only the hit + per-line caps are unit-asserted.

## Base
Branched off `origin/main` `d8ddf4e8`; `origin/main` advanced to `d8792e70` (#588, friday-storage Debug-redaction, non-overlapping) during the work, so this is **rebased onto `d8792e70`**. `git diff --name-only origin/main..HEAD` = the 3 files above.

## Local gates (actual)
- `cargo fmt --check`: clean
- `cargo clippy -p friday-fs -p friday-hub --all-targets -- -D warnings`: clean
- `cargo test -p friday-fs -p friday-hub`: **0 failures** (friday-fs search_fs.rs 11/11; friday-hub lib 303/0).

### DIFF
--- a/rust-core/crates/friday-fs/src/lib.rs
+++ b/rust-core/crates/friday-fs/src/lib.rs
@@ -675,6 +675,227 @@ pub fn edit_file_within_root(
     write_file_within_root(root, candidate, updated.as_bytes())
 }
 
+/// A single line that matched a [`search_within_root`] query.
+///
+/// `relative_path` is the path of the matching file **relative to the root**, with `/`
+/// separators (never absolute, never containing `..` — it is built only from `readdir`
+/// entry names under the contained root). `line_number` is 1-based. `line_text` is the
+/// matching line with its terminator stripped, **truncated to at most
+/// [`SEARCH_MAX_LINE_BYTES`] bytes on a UTF-8 char boundary** (so a single pathological
+/// long line can never blow the output bound).
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub struct SearchHit {
+    /// Path of the matching file, relative to the root (`/`-separated, no `..`/absolute).
+    pub relative_path: String,
+    /// 1-based line number of the match within the file.
+    pub line_number: u64,
+    /// The matching line (terminator stripped), truncated to [`SEARCH_MAX_LINE_BYTES`].
+    pub line_text: String,
+}
+
+/// Hard cap on the total number of [`SearchHit`]s a single search returns. The walk stops
+/// the instant this is reached, so output (and the `Vec` backing it) is bounded regardless
+/// of how many lines match across the tree.
+pub const SEARCH_MAX_HITS: usize = 200;
+/// Hard cap on the byte length of each [`SearchHit::line_text`] (truncated on a char
+/// boundary). Defends against a single multi-megabyte line.
+pub const SEARCH_MAX_LINE_BYTES: usize = 512;
+/// Hard cap on the number of files OPENED+scanned in a single search. Once reached the walk
+/// stops, so a tree with millions of files cannot make the search unbounded.
+pub const SEARCH_MAX_FILES: usize = 2_000;
+/// Hard cap on the number of directories DESCENDED into in a single search. Bounds traversal
+/// over a pathologically wide/deep tree.
+pub const SEARCH_MAX_DIRS: usize = 4_096;
+/// Hard cap on the bytes read from any ONE file (via `Read::take`, so it holds even if the
+/// file grows after we stat it — no TOCTOU on the size). Bounds per-file memory; bytes past
+/// this point are not scanned.
+pub const SEARCH_MAX_FILE_BYTES: u64 = 1_048_576; // 1 MiB
+
+/// READ-ONLY literal-substring search over the UTF-8 text files contained in `root`,
+/// recursive from the root (or, when `subpath` is `Some`, from that contained sub-directory
+/// or single file).
+///
+/// # Containment (no new attack surface)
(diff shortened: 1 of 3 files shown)


thesongzhu:
Addressed the MEDIUM review finding (uncapped descent worklist; false "never unbounded memory" doc). Containment/symlink/resolve logic was SECURE and is untouched.

**Worklist cap.** Every enqueue is now guarded by a `dirs_enqueued < SEARCH_MAX_DIRS` budget (centralized in a `push_dir` closure, applied to the initial push too), so `worklist.len() <= dirs_enqueued <= SEARCH_MAX_DIRS` holds at all times — the pending worklist can no longer accumulate the full cross-directory fanout of the tree; peak worklist memory is now a constant × max-path-len. This is **bounded descent**: search still visits up to `SEARCH_MAX_DIRS` directories (visit-count semantics unchanged from before — the old pop-time check already capped visits; this fix only bounds *peak worklist memory*). Directories beyond the budget are simply not enqueued. The now-subsumed pop-time `dirs_visited` check is kept as a commented defensive floor.

**Honest `# Bounds` doc.** Rewritten to enumerate what IS bounded (hits, per-line bytes, per-file bytes via `Read::take`, files opened, dirs listed, AND the pending worklist) and to document the ONE remaining **inherited** residual: a single directory's entry listing is materialized in full by the shared `list_dir_within_root` primitive (pre-existing; the IDENTICAL allocation the `list_dir` tool makes — search adds no new listing path), so peak memory for one listing scales with that directory's entry count. Deliberately left to the shared primitive (touching it is riskier than documenting the inherited residual); mitigation is operational — host workspaces should not contain adversarial mega-directories.

**Tests** (`tests/search_fs.rs`, all deterministic, suite runs in <1s): `SEARCH_MAX_DIRS` (wide flat tree past the cap — a sentinel match in a dropped, highest-sorted dir is NOT searched while an early control IS, proving the enqueue cap held), `SEARCH_MAX_FILES` (a sentinel sorted past the file cap is never opened), `SEARCH_MAX_FILE_BYTES` (a match positioned beyond the per-file cap is NOT returned — `Read::take` truly bounds the read — while a within-cap match is).

Local gate (actual): `cargo fmt --check` clean; `cargo clippy -p friday-fs -p friday-hub --all-targets -- -D warnings` clean; `cargo test -p friday-fs -p friday-hub` = 388 passed / 0 failed. `git diff --name-only origin/main..HEAD` = the three search files (the hub file carries only the original search wiring from 95baea70 — unchanged by this fix).

PR #590, opened 2026-06-08
feat(rust-fs): hardened shell-free run_command tool (gate→Ed25519-approval gated)

thesongzhu:
## What

Implements the **highest-risk** fs tool, `run_command` (shell-exec), behind the **EXISTING** gate/approval seam. Adds the friday-fs primitive + the friday-hub executor arm. Nothing else.

## Truth label

- `run_command` is **mutating + `Risk::High`** in the trusted registry (already registered; this PR does NOT re-register).
- It routes **gate → `RequiresApproval` → operator Ed25519-signed single-use approval** before executing **ONE** command — the exact command bound into `action_digest` (`resume.rs` re-derives + cross-checks the digest before Ed25519 verify, so a param/command swap → `DigestMismatch` → Deny).
- A command containing shell metacharacters (`;|&\`$(){}` / control chars) is already classified **Critical → gate DENY** by `shell_risk` — it never reaches this arm.
- **PROOF-ONLY, NOT a v1 GO.** `executeRun` is **NOT** replaced. The capability is **DORMANT** until an operator signs a command; there is no auto-execution path.

## Security model (as implemented in `friday_fs::run_command_in_root`)

1. **NO SHELL.** Minimal quote-aware tokenizer (single/double quotes for args-with-spaces; no expansion, no globbing, no `$VAR`, no escape processing) → **direct argv exec** (`program=argv[0]`). No `sh -c`. A slipped-through metachar is an inert literal arg. `echo $HOME` prints the literal `$HOME`.
2. **ENV SCRUB.** `env_clear()` then a **fixed allow-list only**: `PATH=/usr/bin:/bin:/usr/local/bin`, `LANG=C`, `LC_ALL=C`. The child cannot read `FRIDAY_DEEPSEEK_API_KEY` or any inherited secret from `env` (test asserts the child env is *exactly* the allow-list — no `HOME`, no `FRIDAY_*`).
3. **CWD CONTAINMENT.** Per-child `current_dir(canonicalized root)`; never global `set_current_dir`. Missing root → `NotFound`.
4. **TIMEOUT + KILL.** 30s wall-clock deadline (`RUN_COMMAND_TIMEOUT`). Child `setpgid`'d into its own group and **SIGKILL'd (killpg) on every return path** — including a fast-exiting command that left a grandchild holding the pipe write-fd, so the reader-join cannot hang unbounded. Capture is **deadlock-free**: reader threads drain stdout+stderr to EOF, so a child exceeding the OS pipe buffer can never block on write. (Residual sub-ms pgid-reuse window documented in-code; accepted over an unbounded hang — the same trade `std::process::Command::output()` makes.)
5. **OUTPUT BOUND.** Combined stdout+stderr capped at **64 KiB retained** (`RUN_COMMAND_MAX_OUTPUT_BYTES`, char-boundary-safe); `output_truncated` set; bytes past the cap are read-and-discarded. Further bounded to 2048 by the model-feedback path downstream.
6. **FAIL CLOSED.** Tokenize failure → `FsError::CommandInvalid`; spawn failure → `FsError::CommandSpawn`. Both **static messages** — no path/command/secret/errno. A successful spawn always returns `Ok` (incl. non-zero exit and timeout — those are results, not errors).

## Hub arm (`FsToolExecutor::execute`)

- `content` (model-facing) = bounded output + an `[run_command: exit … / timed out]` status line.
- `summary` (**REFS-ONLY → hash-chained audit ledger**) = `run_command: exit {:?}, {n} bytes [(truncated)][(timed out)]` — **NEVER** the output text, **NEVER** the command string (the command already lives in the audit action label + digest), **NEVER** any env/secret. A test asserts the summary contains neither the output nor the command.

## Tests

friday-fs (helper, deterministic/fast): tokenizer, simple `echo`, **cwd-is-root** (`ls` sees only root, not a sibling outside), **env scrubbed to exact allow-list**, **no-shell** (`$HOME` literal), **timeout** (`sleep 5` under 200ms → `timed_out`, returns < 3s), **output bound** (`seq 1 200000` → truncated, retained ≤ cap), non-zero exit is a result, unparseable/unknown-program/missing-root fail closed.

friday-hub: direct-exec arm runs + **summary refs-only** (asserts no output/command leak); **gate-PAUSE** — `run_command` (`touch SENTINEL_RAN`, metachar-free → stays High) PAUSES under deny-all with `calls==0` and no side-effect file (added to `each_mutating_fs_tool_pauses_under_deny_all_and_never_executes`); repurposed the old `executor_run_command_is_unsupported` test to cover the fail-closed `other =>` catch-all via a bogus action.

## Scope

`git diff --name-only origin/main..HEAD`:
- `rust-core/crates/friday-fs/src/lib.rs`
- `rust-core/crates/friday-hub/src/lib.rs`

Does **NOT** touch `gate.rs`, `tool_policy.rs`, `resume.rs`, the registry, or the digest/approval logic — they are already correct; this rides them.

## Local gate (actual results)

- `cargo fmt --check` — clean
- `cargo clippy -p friday-fs -p friday-hub --all-targets -- -D warnings` — exit 0, no warnings
- `cargo test -p friday-fs -p friday-hub` — **0 failures** (friday-hub lib 304 passed; friday-fs lib 17 + adverse 25 + search 14 + primitives 14 passed; arch boundary tests pass)

### DIFF
--- a/rust-core/crates/friday-fs/src/lib.rs
+++ b/rust-core/crates/friday-fs/src/lib.rs
@@ -160,6 +160,21 @@ pub enum FsError {
     /// An underlying I/O error not classified above.
     #[error("io error: {0}")]
     Io(#[from] std::io::Error),
+
+    /// `run_command_in_root` could not turn the `command` string into an argv: it was
+    /// empty/whitespace-only, or had an unbalanced quote. **Static message — carries NO
+    /// path, NO command text, NO secret** (the offending string never reaches the error,
+    /// so it cannot leak into a `Display` that flows to a log/ledger). Fail-closed: an
+    /// unparseable command is refused, never silently run.
+    #[error("command could not be parsed into an argument vector")]
+    CommandInvalid,
+
+    /// `run_command_in_root` could not spawn the child (program not found on the fixed
+    /// child `PATH`, not executable, or the OS refused the exec). **Static message — the
+    /// underlying `io::Error` (which can carry the program name / errno text) is
+    /// DELIBERATELY dropped** so no path/secret reaches the error. Fail-closed.
+    #[error("command could not be spawned")]
+    CommandSpawn,
 }
 
 /// The kind of a [`stat_file_within_root`] target, after the no-follow `lstat`. A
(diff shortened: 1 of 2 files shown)


thesongzhu:
## Fix: deadline-bounded output drain (setsid-escape hub-thread DoS)

Addresses the HIGH reliability defect in `run_command_in_root_with_timeout`.

### The defect (empirically reproduced at the prior head)
The wall-clock timeout only guarded `rx.recv_timeout` (waiting for the **leader**). After it returned, the stdout/stderr reader threads were joined with **no deadline** (`r.join()` over a `drain_into` that blocks in `read()` until EOF). A grandchild that calls `setsid()`/`setpgid(0,0)` **escapes** the leader's process group, so the unconditional `killpg(-pid)` never reaches it; if it inherited and **keeps the pipe write-fd open**, the pipe never EOFs → `drain_into` blocks forever → the unbounded join **pins the calling hub agent-loop thread**. The 30s timeout gave **zero** protection: `recv_timeout` returns the instant the leader exits (`timed_out=false`), falling straight into the unbounded join. Standard double-fork/`setsid` daemonization (retaining stdout) triggered it. The prior module + inline comments **falsely** claimed killpg prevented this.

Repro: a `perl` leader forking a `POSIX::setsid()` grandchild that holds stdout and sleeps 10s, then exits → the call returned only after the **grandchild's full 10s lifetime** (`timed_out=false`).

### The fix (mechanism: poll-deadline, no thread leak)
Scope: **`friday-fs/src/lib.rs` only**. Gate/approval, env-scrub, no-shell, cwd, audit-refs-only and tokenizer are **unchanged** (all verified secure).

- One absolute `drain_deadline = now + timeout + RUN_COMMAND_DRAIN_GRACE` (new const, grace = 2s, for in-group flush after the leader exits).
- `drain_into` now sets the pipe read-fd **non-blocking** (`fcntl O_NONBLOCK`) and `poll()`s against the **remaining** budget. EOF/`POLLHUP` still wakes it immediately, so normal `echo` and the **in-group `sleep 5` timeout-kill still return fast (~200ms)** — only a never-EOF escaped grandchild drains to the deadline. The read is attempted unconditionally on `poll > 0` (gating on `POLLIN` would busy-spin on `POLLHUP`); `EAGAIN`/`EINTR` re-poll.
- At the deadline the reader **marks the output truncated and returns**, so the threads self-terminate and the **unchanged** join loop is naturally bounded — **no unbounded block, no detached-thread leak**. The call **always** returns within ~`timeout + grace`.

### Honest orphan residual (corrected comments)
A deliberately process-group-escaping (`setsid`) grandchild **cannot** be reaped by a process-group kill — that is **inherent** to approved shell-exec: an operator who signs a daemon-spawning command gets a daemon. We do not hunt it down. The corrected module (§4) and inline comments now state this honestly: in-group descendants are SIGKILL'd via killpg; a setsid-escaping grandchild is **not** reaped (inherent), **but** the reader drain is deadline-bounded so the **hub thread is never pinned**.

### Test
`run_command_setsid_escaped_grandchild_does_not_pin_thread`: a perl leader forks a setsid grandchild holding stdout + sleeping 10s, then exits; asserts the call returns in `< timeout+3s` (1s timeout ⇒ `< 4s`), not the grandchild's lifetime, with `output_truncated` set.
- **Verified FAILS pre-fix**: returned in **10.012s** (the full grandchild lifetime), blowing the 4s bound.
- **PASSES post-fix**: ~3.0s, stable over repeated runs.
- All existing run_command tests still pass (echo, cwd, env-scrub allowlist, no-shell, in-group timeout-kill, output cap, tokenizer).

### Local gate (all green)
- `cargo fmt --check` ✓
- `cargo clippy -p friday-fs -p friday-hub --all-targets -- -D warnings` ✓ (0 warnings)
- `cargo test -p friday-fs -p friday-hub` ✓ (0 failures)

Diff scope unchanged: `friday-fs/src/lib.rs` + the unchanged hub file from the original commit.

thesongzhu:
Amended (force-push `d9c72124` → `3fd1436b`, same 2-file diff scope) to fold in two non-behavioral follow-ups from review:

- **Doc correction:** the `CommandRunResult.output_truncated` field docs (struct-level and field-level) said cap-only; this change adds a *second* trigger (the drain deadline cutting reading short before EOF). Both doc lines now state `output_truncated` ⇒ retained output is **incomplete** — either past the byte cap **or** cut by the drain deadline. (The hub already surfaces it as "(truncated)" = incomplete, which is correct for both.)
- **Test robustness:** widened the setsid test's grandchild sleep 10s → 30s and the assertion bound 4s → 8s. The drain returns at ~3.0s (1s timeout + 2s grace) in every run; the wider bound gives generous headroom for loaded-CI scheduling jitter while staying far below the 30s grandchild lifetime, so it still cleanly distinguishes "deadline-bounded" from "pinned". Still fails pre-fix (~30s ≫ 8s), passes post-fix (~3.0s).

Gate re-run green: `cargo fmt --check`; `cargo clippy -p friday-fs -p friday-hub --all-targets -- -D warnings`; `cargo test -p friday-fs -p friday-hub` (0 failures); `cargo doc -p friday-fs --no-deps` (intra-doc links resolve).

PR #591, opened 2026-06-08
feat(rust-hub): read-only hub_providers_detect surface (refs-only, no secrets/network/quota)

thesongzhu:
## What

Adds a new read-only bin `rust-core/crates/friday-hub/src/bin/hub_providers_detect.rs` that surfaces the existing `friday_providers::detect` engine as a reachable, **refs-only** entrypoint.

- Args: `--probe codex|claude|both` (default `both`).
- Builds the production `CliProbe` and calls `friday_providers::detect(&probe, provider)` for each requested provider.
- Emits ONE refs-only JSON object to stdout: `{ "truth_label": "rust_providers_detect", "proof_only": true, "ok": true, "detected": [ { "provider", "installed", "authenticated", "detail" } ] }`.

## Truth label — read-only providers-detect surface

It runs ONLY each provider CLI's official, read-only **status** command (`codex login status`, `claude auth status`) — never a prompt/send. So:

- **No model call** occurs.
- **No quota** is spent.
- **No network** is touched beyond the CLI's own local auth check.
- **No secret** is read.

This resolves the (now 503) `providers/detect` onboarding surface **at the Rust layer**.

## Secret hygiene

The bin emits ONLY the safe `ProviderAuthStatus` fields (provider label, `installed`/`authenticated` booleans, and a coarse static `detail` ∈ `logged_in | not_logged_in | not_installed`). It **NEVER** emits the raw `ProbeOutput` (CLI stdout/stderr), which can carry `authMethod` / `subscriptionType` / account identifiers:

- Primary defense: `parse_status` discards the raw `ProbeOutput` and returns only booleans + a `&'static str` label.
- Structural guarantee: `ProbeOutput` is not `Serialize`, so accidental serialization is a **compile error**, not a runtime leak.
- Backstop: a `reject_forbidden_output` guard (mirroring `hub_run_task`, plus the raw-CLI-field markers `authMethod`/`subscriptionType`/`loggedIn`) runs before printing; on trip → fail-closed `output_guard`, exit 2.
- Fail-closed `BridgeError { kind: &'static str }` for any error (no path/detail/secret leak). Exit 0 on success, 2 on error.

## Tests (mock `ProviderProbe`)

- exact safe JSON shape (4 keys per entry, no raw fields);
- `not_installed` is surfaced, never a fallback;
- parsed surface **strips** raw account fields even when the mock returns `authMethod`/`subscriptionType`/email-laden stdout;
- guard **canary** — a raw `authMethod`/secret marker trips the guard directly;
- guard **passes** on the bin's actual rendered output for every probe selection (no marker collision);
- never serializes raw `ProbeOutput` (`stdout`/`stderr` absent).

## CI

The `rust-core.yml` workflow has no single `--bin` list; it has one explicit per-bin build step each (`hub_run_task`, `hub_run_readback`). I added a **third** mirroring step (`cargo build -p friday-hub --bin hub_providers_detect`) so a compile-broken bin reds CI loudly. Note: `clippy --workspace --all-targets` and `test --workspace` already compile every bin, so this step is loud-fail/parity convention, not strictly required for the bin to build.

## Local gate (actual results)

- `cargo fmt --check`: clean
- `cargo clippy -p friday-hub --all-targets -- -D warnings`: clean
- `cargo build -p friday-hub --bin hub_providers_detect`: ok
- `cargo test -p friday-hub`: 0 failures (7 new bin tests pass; full suite green)
- Local run `--probe both` (both CLIs present + logged in on this host): `{"detected":[{"authenticated":true,"detail":"logged_in","installed":true,"provider":"codex"},{"authenticated":true,"detail":"logged_in","installed":true,"provider":"claude"}],"ok":true,"proof_only":true,"truth_label":"rust_providers_detect"}` — refs-only (booleans + labels only, no account info). Bad-arg → `{"error_kind":"bad_args",...}` exit 2.

## Scope

Exactly two paths: the new bin + the CI yml step. No lib.rs run-loop change, no friday-fs, no friday-providers internals, no TS, no gate/schema.

## Honest framing

**PROOF-ONLY.** Does NOT register a production route, does NOT replace `executeRun`, and confers **no v1 GO**. TS wiring for `/v1/providers/detect` is deferred (design-handoff gated) and out of scope here — the bin is the deliverable.

### DIFF
--- a/.github/workflows/rust-core.yml
+++ b/.github/workflows/rust-core.yml
@@ -48,6 +48,13 @@ jobs:
         # a missing/compile-broken `hub_run_readback` reds CI loudly. Only BUILT here — the
         # TS bridge test spawns a scripted MOCK bin instead (no DB, no key, no quota).
         run: cargo build -p friday-hub --bin hub_run_readback
+      - name: Build bins (incl. read-only hub_providers_detect surface)
+        # Read-only providers-detect surface over friday_providers::detect: names the bin
+        # explicitly so a missing/compile-broken `hub_providers_detect` reds CI loudly. Only
+        # BUILT here — running it invokes ONLY each provider CLI's local read-only `status`
+        # command (no model call, no quota, no network beyond the CLI's own auth check, no
+        # secret), so it is safe but unnecessary to execute in CI.
+        run: cargo build -p friday-hub --bin hub_providers_detect
       - name: Test
         # The one live DeepSeek test is #[ignore]d, so this needs no network/secrets.
         run: cargo test --workspace
--- a/rust-core/crates/friday-hub/src/bin/hub_providers_detect.rs
+++ b/rust-core/crates/friday-hub/src/bin/hub_providers_detect.rs
@@ -1,0 +1,368 @@
+//! Read-only providers-detect surface — `hub_providers_detect`.
+//!
+//! PROOF-ONLY. A thin one-shot bin that surfaces the existing
+//! [`friday_providers::detect`] engine as a reachable, refs-only entrypoint,
+//! cloning the read-only output-guard shape of the `hub_run_task` write-bridge.
+//!
+//! It runs ONLY each provider CLI's official, read-only **status** command
+//! (`codex login status`, `claude auth status`) via [`friday_providers::CliProbe`]
+//! — never a prompt/send — so **no model call** occurs, no quota is spent, no
+//! network is touched beyond the CLI's own local auth check, and no secret is read.
+//!
+//! This resolves the (now 503) `providers/detect` onboarding surface at the Rust
+//! layer. It does NOT register a production route and confers no v1 GO. TS wiring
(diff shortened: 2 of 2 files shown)


thesongzhu:
Fixed LOW fail-closed defect: `main()` now reads argv via `env::args_os()` and converts each `OsString -> String` fail-closed (any non-UTF-8 arg in ANY position -> coarse `bad_args`), so `hub_providers_detect $'\xff'` returns refs-only `{"error_kind":"bad_args"}` + exit 2 instead of panicking at exit 101 (verified end-to-end on the built bin). Added `non_utf8_argv_is_fail_closed_not_a_panic` (#[cfg(unix)], bare-positional 0xff, would panic pre-fix); existing 7 tests unchanged and green. Gate local: fmt --check OK, clippy -D warnings OK, build OK, cargo test 8/8 (bin) all green.

PR #595, opened 2026-06-08
refactor(rust-hub): consolidate refs-only output guard into one shared helper

thesongzhu:
## What

Seven friday-hub proof bins each carried their **own copy-pasted** `reject_forbidden_output()` guard — the same secret/path marker list + scan loop duplicated across `hub_run_task`, `hub_run_readback`, `hub_providers_detect`, `diagnostics_snapshot`, `hub_authed_run`, `hub_resume_approval`, and `mission_workbench_projection`. Copy-pasting a **security** guard risks drift: a fix to one not propagating, or a new bin copying a stale list (raised in the #592 adversarial note — low/inert but real maintainability risk).

This consolidates the **common** marker set + scan loop into a single shared module `friday-hub/src/refs_guard.rs`, declared via a 1-line `pub mod refs_guard;` near the top of `lib.rs` (line ~162, far from the FsToolExecutor/run-loop region). Each bin keeps a **3-line typed adapter** that delegates to the shared helper, passing its bin-specific body-field markers as `extra` — so every bin's **exact** blocking set is preserved (just unioned with the broadened common path set), and every call site + existing test stays byte-identical.

```rust
pub fn reject_forbidden_output(rendered: &str, extra: &[&str]) -> Result<(), String>
```
Always checks the common set `{Authorization, Bearer, sk-, /Users/, /private/, /home/, /var/, /tmp/, /etc/}`; additionally checks each `extra` marker. Returns the matched marker on a hit so callers map it into their own error type (mission_workbench keeps its exact `forbidden marker in projection: {marker}` message).

## Hardening (defense-in-depth, inert today)

- **Broaden the common absolute-path markers** from `{/Users/, /private/}` to also cover `{/home/, /var/, /tmp/, /etc/}`. The guard is **strictly stricter** — it can only newly reject, never newly permit.
- **Guard the Err-branch.** Each bin's error-payload stdout print now routes through the same guard (fail closed if a marker ever leaked). `error_kind` is a static closed-vocab token today, so this **never suppresses output** — success and error stdout are byte-identical. (`mission_workbench`'s error path is `eprintln`-only — nothing to guard.)

## Behavior preservation

- No bin's **success-path payload construction** or Ok-branch print is touched (verified: no success-payload field and no `println!("{rendered}")` Ok-branch line appears in the diff).
- Success output is **byte-identical for every guard-passing input.** The only behavior the stricter guard can produce is failing closed on an output that previously passed. Verified no bin's safe output contains any new marker against **static fixtures**. Three bins emit runtime free-form DB text (`mission_workbench` advisorSummary/alternatives; `hub_run_readback` event_kinds; `hub_run_task` executed_tools) — these rest on the **same no-absolute-paths contract** the pre-existing `/Users/`,`/private/` markers already enforced; broadening is consistent with that contract, and failing closed on a leaked absolute path is the intended defense.

## Local gate (exactly what was run)

- `cargo fmt -p friday-hub --check` — clean
- `cargo clippy -p friday-hub --all-targets -- -D warnings` — 0 warnings
- `cargo test -p friday-hub` — **0 failures** (316 lib incl. 6 new `refs_guard` tests; all 6 per-bin `forbidden_output_guard_*` + shape tests pass untouched; 9 integration; live/probe ignored)
- All 7 affected bins build

(Did **not** run the CI `secrets` job locally; `refs_guard.rs` contains the same low-entropy marker literals already present on main in the bin test modules — none a real provider-key shape.)

## Scope

`refs_guard.rs` (new) + the 7 bins + a 4-line `mod` decl in `lib.rs`. No friday-fs, no TS, no run-loop region (away from ~671/1400/2160 to avoid conflict with #590, confirmed non-overlapping). Removes the security-guard drift risk.

PROOF-ONLY maintainability/security hardening — **not a v1 GO**.

### DIFF
--- a/rust-core/crates/friday-hub/src/bin/diagnostics_snapshot.rs
+++ b/rust-core/crates/friday-hub/src/bin/diagnostics_snapshot.rs
@@ -202,9 +202,9 @@ fn arg_value(args: &[String], name: &str) -> Option<String> {
 /// these markers should never appear; the guard is the backstop that fails the
 /// WHOLE projection closed if one ever does (non-zero exit + refs-only error).
 fn reject_forbidden_output(rendered: &str) -> Result<(), BridgeError> {
-    // Delegates to the single shared guard (common secret/path markers, now broadened
-    // to /home,/var,/tmp,/etc). This bin has no extra body-field markers — the payload
-    // carries only counts/booleans/ids/static labels.
+    // Delegates to the single shared guard (common secret/path markers
+    // Authorization/Bearer/sk-/`/Users/`/`/private/`). This bin has no extra body-field
+    // markers — the payload carries only counts/booleans/ids/static labels.
     friday_hub::refs_guard::reject_forbidden_output(rendered, &[])
         .map_err(|_| BridgeError::new("output_guard"))
 }
--- a/rust-core/crates/friday-hub/src/bin/diagnostics_snapshot.rs
+++ b/rust-core/crates/friday-hub/src/bin/diagnostics_snapshot.rs
@@ -83,7 +83,13 @@ fn main() {
                 "ok": false,
                 "error_kind": err.kind,
             });
-            println!("{payload}");
+            // Defense-in-depth: route the error payload through the SAME guard as the
+            // success path (fail closed if a marker ever leaked). `error_kind` is a
+            // static closed-vocab token today, so this never suppresses output.
+            let rendered = payload.to_string();
+            if reject_forbidden_output(&rendered).is_ok() {
+                println!("{rendered}");
+            }
             eprintln!("diagnostics_snapshot_unavailable: {}", err.kind);
             std::process::exit(2);
         }
@@ -196,12 +202,11 @@ fn arg_value(args: &[String], name: &str) -> Option<String> {
 /// these markers should never appear; the guard is the backstop that fails the
 /// WHOLE projection closed if one ever does (non-zero exit + refs-only error).
 fn reject_forbidden_output(rendered: &str) -> Result<(), BridgeError> {
-    for marker in ["Authorization", "Bearer", "sk-", "/Users/", "/private/"] {
-        if rendered.contains(marker) {
-            return Err(BridgeError::new("output_guard"));
-        }
-    }
-    Ok(())
+    // Delegates to the single shared guard (common secret/path markers, now broadened
+    // to /home,/var,/tmp,/etc). This bin has no extra body-field markers — the payload
+    // carries only counts/booleans/ids/static labels.
+    friday_hub::refs_guard::reject_forbidden_output(rendered, &[])
+        .map_err(|_| BridgeError::new("output_guard"))
 }
 
 #[cfg(test)]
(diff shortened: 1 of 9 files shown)


thesongzhu:
## Follow-up fix: dropped the over-blocking `/home,/var,/tmp,/etc` markers (`ea6bc722`)

Adversarial review found a **MEDIUM over-block regression** in the consolidation: alongside moving the per-bin refs-only guard into the shared `refs_guard.rs`, this PR also **broadened** `COMMON_MARKERS` with `/home/`, `/var/`, `/tmp/`, `/etc/` on top of the original `Authorization`/`Bearer`/`sk-`/`/Users/`/`/private/`.

### The defect
Those are bare **substring** matches. A legitimately-contained **relative** tool path with an interior dir segment named `etc`/`var`/`tmp`/`home` (e.g. `config/etc/app.conf`) is surfaced verbatim by `hub_run_readback` inside `tool.executed:` event kinds. So the success payload contains the substring `/etc/` → the shared guard returns `Err` → the bin exits `2` (`output_guard`) and **refuses a readback it previously emitted**. A real fail-closed availability regression — no leak prevented (proof refs are hex fingerprints, not paths). The original per-bin guards never listed these roots, so they passed.

### The fix (scope: `refs_guard.rs` markers + doc + tests; 7 bin comment lines)
- **Markers**: `COMMON_MARKERS` reverted to EXACTLY the pre-#595 common set — `Authorization`, `Bearer`, `sk-`, `/Users/`, `/private/`. Removed `/home,/var,/tmp,/etc`. The **DRY consolidation is kept** (shared helper + per-bin adapters + Err-branch guarding stay). Every bin's blocking set (common ∪ its extras) is now **byte-identical to pre-#595**.
- **Docs**: rewrote the false claim that the broadening "strictly tightens the guard without changing any bin's verified safe output." It's now honest: a **pure DRY consolidation** matching the pre-existing markers; the #592 "broaden markers" suggestion is **refuted** (over-blocks legit relative paths) and recorded for the record. Also corrected the one-line delegation comment in each of the 7 bins (comment-only; behavior unchanged).
- **Tests**:
  - Removed the rows asserting `/home,/var,/tmp,/etc` trip; added `dropped_unix_root_markers_no_longer_trip_as_absolute_paths`.
  - Kept tests asserting each original common marker (`Authorization`/`Bearer`/`sk-`/`/Users/`/`/private/`), extras, and safe refs-only output.
  - **Added regression test** `interior_dir_segment_named_like_a_unix_root_does_not_false_trip`: a readback-style payload with `tool.executed:read 15 bytes from config/etc/app.conf` (+ `var/`, `tmp/` interior segments) now **PASSES** (`Ok`). Verified empirically: this test **FAILS pre-fix** (markers still present → `/etc/` trips) and **PASSES post-fix**.

### Local gate (all green)
- `cargo fmt -p friday-hub --check` ✅
- `cargo clippy -p friday-hub --all-targets -- -D warnings` ✅ (0 warnings)
- `cargo test -p friday-hub` ✅ (318 lib + all bin tests, 0 failures)
- All 7 affected bins build ✅

`git diff --name-only origin/main..HEAD` remains the same 9 files; only `refs_guard.rs` substance changed in this follow-up (the 7 bins are one-line comment corrections, `lib.rs` unchanged).

PR #643, opened 2026-06-10
feat(rust-workflow): R2 — workflow run-control (resume/retry/cancel) Rust ownership (dark, no route flip; resume live, retry/cancel deferred — see body)

thesongzhu:
**1 of 3 run-control ops is functionally LIVE** (resume); retry/cancel are documented, fail-closed deferrals (not stubs). DARK substrate: no production route, no scheduler/trigger/daemon, no migration, no route flip. Changes no TS file. Purely additive over the existing Rust run model. NOT v1 GO.

## What it mirrors
The TS run-control path — `resumeRun` / `retryRun` / `cancelRun` in `src/workflows/services/friday-workflow-execution-service.ts` — is already METHOD-guarded / fail-closed in the live runtime (the **G4 retirement guard**: every non-route caller is fenced out with 503 `TS_RUNTIME_WORKFLOW_RUNS_RETIRED` while runtime ownership moves to Rust). This PR builds the Rust-owned equivalent of that control surface.

**Run-model note (a deliberate divergence the task prompt conflated):** the prompt described the state machine as `Pending → Running → AwaitingCheckpoint → Done/Failed` — but that is the *Rust* `friday-core::WorkflowRunState` (5 states). The *actual* TS `WorkflowRunStatus` (`friday-workflow-run-machine.ts`) is an 8-state machine (`queued/running/pausing/paused/compensating/completed/failed/cancelled`) with a `failed→running` retry edge and `*→cancelled`. The Rust hub does **not** own the TS schema; it owns the `friday-core` 5-state model persisted in `friday_storage::workflow` and driven by the live S9 engine (`workflow_exec`). So this module mirrors the TS control **logic** (validation order, fail-closed error vocabulary, run effects) mapped onto the Rust model (TS `paused` ≈ Rust `AwaitingCheckpoint`) — it does **not** port the 8 TS state-strings into the Rust `state` column (which would silently corrupt every 5-state reader, since `parse_run_state` coerces unknown strings to `Failed`).

## Per-op scope
[...]

PR #645, opened 2026-06-10
feat(rust-workflow): R2 slice-2 — retry/cancel run-control + friday-core Cancelled/retry-edge (additive) + m0027 (dark)

thesongzhu:
## What this does
Completes the **dark** Rust workflow **run-control plane**. R2 slice-1 (#643) shipped `resume()` and returned `retry()`/`cancel()` as `NotRepresentable` no-write deferrals; the R2 review pinned the exact blocker — `friday-core`'s `WorkflowRunState` had no terminal `Cancelled` and no `Failed->Running` edge, and `workflow_step` had no per-attempt / `cancel_reason` columns. This slice makes the **minimal additive** core + storage extension and builds both ops over it. **DARK: no production route, no flag, no scheduler/trigger, no TS file touched.**

## 1. friday-core — the additive core change (and proof of no behavior change)
`crates/friday-core/src/workflow.rs`, **ADDITIVE ONLY**:
- New terminal variant `WorkflowRunState::Cancelled` (distinct from `Failed`).
- New edges: `Failed -> Running` (retry, the only new edge into `Running` — does **not** widen resume) and `{Pending, Running, AwaitingCheckpoint} -> Cancelled` (cancel; terminal `Done`/`Failed`/`Cancelled` have **no** `-> Cancelled` edge → no terminal reopen).
- `as_str()`/`is_terminal()` extended for `Cancelled`.

**No existing variant or edge changed.** The shared enum is consumed by the LIVE S9 engine + `skill.rs`/`planning.rs`/`workflow_read.rs`, but those reference the type only in docs and via `as_str`/`is_terminal`/`matches!` predicates — the **only** exhaustive `match` is `as_str()` in the core file itself (verified by grep). Proof tests:
- `r2_additive_change_preserves_every_pre_existing_edge_and_classification` — every pre-slice-2 edge still ALLOWED, every pre-existing rejection still REJECTED (`Pending->Done`, `Done->Running/Failed`, `Failed->Done`, `Failed->AwaitingCheckpoint`), every label + terminal classification unchanged.
- `r2_cancelled_state_and_retry_edge_are_additive_and_terminal` — `Cancelled` distinct/terminal, new edges allowed, terminal states have no `-> Cancelled` edge, no edge out of `Cancelled`.
- Pre-existing `run_state_machine` is untouched and still passes.

## 2. friday-storage — parse round-trip + m0027 (additive)
- **`parse_run_state`** adds `"cancelled" => Cancelled` — NOT the `_ => Failed` catch-all. This is the **non-compiler-enforced** correctness condition: without it, a cancel write would silently read back as `Failed`, reintroducing the cancelled/failed erasure the hard rules forbid.
- **m0027 (v27)** — two ADDITIVE `ALTER ADD COLUMN`s on the **already-Hub-only** workflow tables: `workflow_step.attempt INTEGER NOT NULL DEFAULT 1` (base attempt = 1, matching TS's 1-based node attempt) and `workflow_run.cancel_reason TEXT` (nullable). **No new table → `HUB_ONLY_TABLES` is correctly untouched** (both columns hang off tables already registered Hub-only). No ALTER that breaks v26 rows.
- New fns: `cancel_run` (atomic terminal `Cancelled` + `cancel_reason`, transition validated through the **same** core machine `set_run_state` uses), `reopen_failed_step` (reopens the non-`Verified` frontier → `Pending`, `attempt += 1`, clears stale evidence_ref; refuses a `Verified` step), `step_attempt` / `run_cancel_reason` readers.
- **KAT** `forward_migration_v26_to_v27_adds_run_control_columns_and_preserves_v26_rows`: seeds a v26 run+step, asserts both columns absent at v26, reopens → v27, asserts pre-existing rows survive with `attempt` backfilled to 1 and `cancel_reason` NULL, then cancels through the typed API and confirms terminal `Cancelled` (NOT Failed) round-trips on the migrated DB.

## 3. friday-hub — retry/cancel over the extended core
- **`retry_workflow`** (new engine entrypoint in `workflow_exec.rs`) mirrors `resume_workflow`'s STRUCTURE — it **reuses** `gate_dispatch` + `advance_from`, does **not** re-implement execution. **Key correctness finding:** the engine **never persists a `Failed` *step* status** — an exec-error on a side-effect step leaves it `ProofPending`, on a pure step `Running`, and a denied checkpoint stays `Pending`. So the retry frontier is the **first non-`Verified`** step (generalizing resume's Pending-only finder). Sequence: precheck run is `Failed` → reopen frontier (`attempt+1`) → `Failed->Running` → manual re-dispatch of the frontier THROUGH THE GATE (no `add_step` → no PK collision) → `advance_from(frontier+1)`. A mutating frontier with no approval **re-pauses** (never executes unapproved). Verified earlier steps are never re-run.
- **`retry()` / `cancel()`** in `workflow_run_control.rs` delegate over the extended core, mirroring `resume()`'s fail-closed precheck pattern (NotFound → InvalidTransition → Definition). `cancel()` never loads a definition or spends executor effort; it writes terminal `Cancelled` + reason and **never coerces onto `Failed`**. `RunControlError::NotRepresentable` removed (now dead).

## retry/cancel semantics vs TS (`friday-workflow-execution-service.ts`)
- `cancelRun(runId, reason)`: TS `assertTransition(status,"cancelled")` + `finalizeRun(...,"cancelled",{message: reason})`. Rust matches: terminal-state refusal + atomic `Cancelled`+reason write. TS also aborts in-flight `AbortController`s; Rust runs are driven synchronously (no background controller), so there is nothing to abort — documented.
- `retryRun(runId, nodeIds)`: TS `assertTransition(status,"running")` + per-node `attempt+1`/idempotency-key retry rows. Rust matches the `Failed->Running` transition + `attempt+1` (in-place UPDATE, since `workflow_step` keys on a single `step_id` PK rather than `(runId,nodeId,attempt)`). Selective-node retry: the linear Rust engine has a single failure frontier, so a supplied `nodes` selection must name exactly that frontier step or it is fail-closed `InvalidTransition` (never silently ignored).

## Deferred ACs (explicit, not stubbed)
- **No idempotency keys** — retry re-drives the frontier (correct for a transient failure); a *partially-applied* side effect could double-run. Documented in the engine + control docs.
- **Arbitrary-subset node retry** — needs a DAG engine; the linear engine has one frontier.

## Test evidence (worktree, base `01e9d591`)
```
cargo build              Finished dev profile
cargo fmt --check        clean (exit 0)
cargo clippy --workspace --all-targets -- -D warnings   Finished (no warnings)

friday-core      test result: ok. 146 passed; 0 failed   (incl. the 2 additive-proof tests)
friday-storage   v26_to_v27 KAT ok · cancel_run_writes... ok · cancel_run_fail_closed... ok · reopen_step... ok
friday-hub       test result: ok. 499 passed; 0 failed
  retry_re_drives_a_real_engine_failure_to_done_through_the_gate ... ok   (transient exec, NO injected state)
  retry_of_a_mutating_frontier_re_pauses_without_an_approval_and_never_executes_unapproved ... ok
  retry_fail_closed_unknown_non_failed_and_bad_node_selection ... ok
  cancel_writes_terminal_cancelled_with_reason_and_never_coerces_to_failed ... ok
  cancel_fail_closed_unknown_and_terminal_runs ... ok
friday-arch-tests  all green
```

DARK substrate — NOT v1 GO.

### DIFF
--- a/rust-core/crates/friday-core/src/workflow.rs
+++ b/rust-core/crates/friday-core/src/workflow.rs
@@ -12,6 +12,15 @@ use crate::error::CoreError;
 
 /// Lifecycle of a workflow run. Checkpoints requiring user action pause the run
 /// in `AwaitingCheckpoint` (those go to Activity, `08` §1/§4).
+///
+/// R2 slice-2 (additive): `Cancelled` is a NEW terminal state (mirrors the TS
+/// `WorkflowRunStatus` `cancelled`, distinct from `failed`), and a `Failed ->
+/// Running` retry edge + `{Pending,Running,AwaitingCheckpoint} -> Cancelled`
+/// cancel edges are added. NO existing variant is changed/removed and NO existing
+/// transition edge is altered — `Pending`/`Running`/`AwaitingCheckpoint`/`Done`/
+/// `Failed` and every prior edge behave byte-identically (the LIVE S9 engine,
+/// `skill.rs`/`planning.rs`/`workflow_read.rs` only reference the type in docs and
+/// `as_str`/`is_terminal` predicates, none of which change for the old variants).
 #[derive(Clone, Copy, Debug, PartialEq, Eq)]
 pub enum WorkflowRunState {
     Pending,
@@ -19,6 +28,12 @@ pub enum WorkflowRunState {
     AwaitingCheckpoint,
     Done,
     Failed,
+    /// Terminal: the run was cancelled by the owner/operator (distinct from
+    /// `Failed` — a cancel is an intentional stop, not an error). NEW in R2
+    /// slice-2; mirrors TS `cancelled`. Coercing a cancel onto `Failed` would
+    /// erase this distinction (the data-fudge the hard rules forbid), so the
+    /// state is first-class.
+    Cancelled,
 }
 
 impl WorkflowRunState {
@@ -29,17 +44,24 @@ impl WorkflowRunState {
             WorkflowRunState::AwaitingCheckpoint => "awaiting_checkpoint",
             WorkflowRunState::Done => "done",
             WorkflowRunState::Failed => "failed",
+            WorkflowRunState::Cancelled => "cancelled",
         }
     }
 
     pub fn is_terminal(&self) -> bool {
-        matches!(self, WorkflowRunState::Done | WorkflowRunState::Failed)
+        // `Cancelled` is terminal alongside `Done`/`Failed` (a cancelled run is a
+        // finished run). The existing `Done`/`Failed` classification is unchanged.
+        matches!(
+            self,
+            WorkflowRunState::Done | WorkflowRunState::Failed | WorkflowRunState::Cancelled
+        )
     }
 
     pub fn can_transition_to(&self, next: WorkflowRunState) -> bool {
         use WorkflowRunState::*;
         matches!(
             (self, next),
+            // --- Pre-R2 edges (UNCHANGED — byte-identical to slice-1) ---
             (Pending, Running)
                 | (Pending, Failed)
                 | (Running, AwaitingCheckpoint)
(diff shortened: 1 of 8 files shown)


thesongzhu:
Post-review follow-up (addressed before merge):
- **Cancelled-run READBACK path verified.** `workflow_read.rs` re-validates only the *step* status (against `StepStatus`, unchanged here) and the step_id shape; the *run* `state` is a verbatim pass-through string with no closed-vocab allowlist, so a cancelled run projects `state = "cancelled"` correctly (no coercion to failed on the read side). Added `run_summary_projects_the_cancelled_state_faithfully_not_coerced_to_failed` to lock this contract.
- **Stale truth-labels in the module header fixed** (no-false-info): the header said "NO migration" (now false — m0027 lands) and described a "5-state … coerces unknown to Failed" enum (now 6-state; `cancelled` is a KNOWN string that round-trips). Both corrected.

PR #661, opened 2026-06-11
fix(ns-hardening): close dot-join namespace collision (TS+Rust shared, F5.5) — TS LIVE / Rust DARK

thesongzhu:
### What this closes
The session-memory namespace is built by `.`-joining 7 segments
(`tenant.<account>.channel.<channel>.user.<user>.shared`) into a composite
string that becomes the memory **`principal_id` SCOPE** (Rust
`lib.rs:6161`, `memory_extraction.rs:456`). The segment normalizer **kept
the literal `.`** — TS `normalizeNamespaceSegment` regex `/[^a-z0-9._-]/g`
(`.` in the keep-class) and Rust `is_kept` returning `true` for `.`.

Because `.` was both a kept intra-segment char **and** the segment joiner,
a userId / accountId / channel literally containing the joiner word (e.g. a
userId `x.user.y`) could **forge a different (account, channel, user)
tuple's** namespace string → a cross-scope memory read/write collision.
Concretely, pre-fix both of these DISTINCT tuples produced the SAME string
`tenant.a.channel.b.user.x.user.y.shared`:
- `(a, b, "x.user.y")`
- `(a, "b.user.x", y)`

### Fix — drop-dot (chosen over escape-dot)
Drop `.` from the segment keep-set in **both** halves identically — `.` now
maps to `-` like any other punctuation. With no `.` inside any segment, the
composite always splits into exactly the seven fixed-position parts, so the
mapping is **injective over normalized segment tuples**: this closes the
**cross-position joiner-injection** (a segment can no longer forge a
different account/channel/user split). The **within-position** normalization
stays lossy (`a.b` ≡ `a-b`, the same accepted behavior as the pre-existing
`@` → `-`) — lower-severity and not exploitable under single-owner v1; this
is NOT a universal-injectivity claim.

This is a one-character, **byte-identical** TS↔Rust change that preserves
the parity contract. Escape-dot was rejected because the TS-regex /
Rust-char-loop escaping schemes diverge and the escape char itself needs
escaping — exactly what the byte-identical parity rule punishes.

A shared collision test (two distinct tuples must never produce the same
namespace + 7-part injectivity assertions) is added in both TS and Rust, and
the stale dotted-segment assertions + doc comments are updated (the residual
the Rust header previously documented as OPEN is now CLOSED).

### Status
- **Rust half = DARK + DEPLOY-GO-gated** — deploying the binary changes no
  live behavior (the Rust extraction path is dark; this is a parity port).
- **TS half = LIVE** — `normalizeNamespaceSegment` is on the live TS
  memory-namespace path. Deploying it **re-scopes** any EXISTING memory whose
  segment legitimately contained a `.` (an email-shaped userId, a dotted
  accountId/channel): that memory's `principal_id` derivation changes, so it
  is orphaned. Under the single-owner v1 threat model (enforce_single_peer,
  one configured owner) userIds are not adversarially controlled, so the
  collision itself is **defense-in-depth**, not an active exploit — but the
  re-scope is a real one-way data effect (see operator gate below).
- **No migration** (string-derivation change only; no schema). No backfill /
  dual-read mitigation is wired (the spec chose migration-free).

### Operator gate (live re-scope)
Before the **TS half** ships, the operator must confirm or accept that any
existing memory in the **prod store** whose `principal_id` has a dotted
segment (email-shaped userId, dotted account/channel) will be orphaned. The
blast radius cannot be verified from this repo — the audit found zero dotted
namespaces in *test expectations*, but the live risk is in the prod memory
store, which is not (and should not be) inspected here. Re-scoping is one-way
and there is no migration/backfill hook.

### What remains
- The operator confirmation/accept above (the only gate).
- This PR is intentionally SEPARATE (disjoint subsystem from the agent-run /
  workflow planes); the TS+Rust edits are in the SAME PR by parity-lockstep
  mandate.

### Tests (honest counts)
- TS: `friday-session-memory-namespace.test.ts` 15 tests pass (incl. new
  collision test); `friday-session-service.test.ts` 95 pass; memory-guard /
  agent-tools / session-routes consumer suites 153 pass. TS lint 0 errors.
- Rust: `session_namespace` lib unit tests 20/20 pass (incl. new
  `dot_join_collision_is_closed_distinct_tuples_never_collide`); full
  `friday-hub` lib 504/504 pass; `memory_extraction` 14, `hub_extract_memory`
  bin 6 pass. `cargo clippy -p friday-hub --all-targets -D warnings` clean (0).
  `cargo fmt` (pinned 1.95.0) applied.

### DIFF
--- a/rust-core/crates/friday-hub/src/session_namespace.rs
+++ b/rust-core/crates/friday-hub/src/session_namespace.rs
@@ -7,8 +7,9 @@
 //! user's memory is isolated PER (account, channel) instead of being shared across every
 //! channel for the same user.
 //!
-//! Format (byte-identical to the TS — including the Unicode-aware lowercasing; see the
-//! normalization note on the dot-join residual that is SHARED with the TS oracle):
+//! Format (byte-identical to the TS — including the Unicode-aware lowercasing; the dot-join
+//! collision is now CLOSED in BOTH halves by dropping `.` from the segment keep-set, see the
+//! normalization note on [`is_kept`]):
 //! ```text
 //! tenant.<accountId>.channel.<channel>.user.<normalizedUserId>.shared
 //! ```
@@ -284,7 +285,7 @@ fn falsy_or<'a>(value: Option<&'a str>, fallback: &'a str) -> &'a str {
 /// Byte-identical port of the TS `normalizeNamespaceSegment`:
 /// ```js
 /// value.toLowerCase()
-///   .replace(/[^a-z0-9._-]/g, "-")
+///   .replace(/[^a-z0-9_-]/g, "-")
 ///   .replace(/-+/g, "-")
 ///   .replace(/^-|-$/g, "");
 /// // empty result -> "default"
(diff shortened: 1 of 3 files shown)


thesongzhu:
QUEUED — DO NOT MERGE (autonomous coordinator, operator asleep). Adversarial review 2/3 BLOCK: this auto-activates a LIVE one-way re-scope of memory_items (namespace re-key) on deploy with NO code-level gate = a real data downgrade, violating the DARK + no-downgrade bar. Needs a design decision (flag-gate / dual-read / backfill) before it can be safe. Flagged for operator wake-review.

thesongzhu:
Superseded by #667. #661 deploys a LIVE one-way memory re-key (dropping `.` from the namespace keep-set orphans any existing memory whose segment legitimately contained a `.`). #667 builds the safe version per the operator's decision: the dot-drop is FLAG-GATED (DEFAULT-OFF, `FRIDAY_NS_HARDENING_ENABLED`) so flag-off is byte-identical with zero re-scope, and the READ path DUAL-READs both the hardened and legacy namespaces so nothing is lost on flip. The additive collision-prevention is kept. Please CLOSE #661 in favor of #667.

thesongzhu:
Superseded by #667 (flag-gate+dual-read redesign per operator). #667 itself queued pending a flag-off recall-regression fix — see #667.

PR #667, opened 2026-06-11
fix(ns-hardening): flag-gate dot-join collision fix DEFAULT-OFF + DUAL-READ (supersedes #661)

thesongzhu:
## ns-hardening REDESIGN: flag-gate the dot-join collision fix DEFAULT-OFF + DUAL-READ

**Supersedes #661 — #661 should be CLOSED in favor of this PR (do-not-merge).**

### Why #661 is unsafe
#661 drops `.` from the namespace segment keep-set unconditionally. The session-memory
namespace (`tenant.<account>.channel.<channel>.user.<user>.shared`) is the memory store
SCOPE. On the **LIVE TS path**, deploying that change re-derives the namespace for any
EXISTING memory whose segment legitimately contained a `.` (an email-shaped userId, a
dotted account/channel) → that memory's scope changes → it is **orphaned**. That is a
one-way destructive re-key (a data downgrade) on deploy, with no migration/backfill.

### What this builds (per the operator's decision)
**(a) FLAG-GATED, DEFAULT-OFF.** The hardened (dot-drop) derivation is governed by
`FRIDAY_NS_HARDENING_ENABLED` (exactly `"1"` = on; unset/empty/`"0"`/other = off — same
convention as `FRIDAY_CLAUDE_ROUTE_ENABLED`). Flag **OFF ⇒ the keep-set KEEPS `.`** exactly
as before ⇒ every namespace is **byte-identical** to the legacy derivation ⇒ **zero
re-scope on deploy**.

**(b) DUAL-READ.** The READ (recall) path resolves an ORDERED, DEDUPED candidate list —
`[legacy]` when off, dedup `[hardened, legacy]` when on — and unions the recall over BOTH,
so memory written under the legacy namespace is STILL recalled after a flag flip. **No
one-way destructive re-key; no migration/backfill.** Dedup-collapse: when no segment carries
a `.`, hardened == legacy, so the list collapses to one entry → zero extra reads in the
common case even with the flag ON.

**(c) ADDITIVE collision-prevention kept.** The hardened derivation closes the cross-tuple
dot-join collision for NEW (hardened) writes — injective over normalized tuples.

### TS (LIVE recall path)
- `resolveFridaySessionMemoryNamespace` gains a `hardened` mode (reads the flag);
  `resolveFridaySessionMemoryNamespaceCandidates` returns the dual-read list;
  `isFridayNamespaceHardeningEnabled` reads the default-off flag.
- `sessionService.getSessionMemoryNamespaceCandidates` **re-DERIVES** both candidates from
  the session axes (NOT the cached write-time `memory_namespace`), so a session created
  before the flip is still recalled via the legacy candidate (+ a defensive stale-cache tail).
- The live recall consumer (`friday-agent-memory-tools.search`) unions the legacy candidate
  tail (both semantic search AND the lexical fallback) and dedups by id.
  `memoryService.search` stays single-namespace per call.

### Rust (DARK + DEPLOY-GO-gated)
- `session_namespace`: `ns_hardening_enabled` (default-off flag), the
  `hardened`-parameterized derivation, and `resolve_session_memory_namespace_candidates`.
- `friday-storage::memory::recall_confirmed_multi`: union + dedup-by-`memory_id`; the
  per-principal SQL stays single-principal (`debug_assert` intact) so cross-principal
  isolation can never widen.

### HONEST scope of the collision fix (documented, not blind-fixed)
The **LEGACY namespace IS the colliding string**, so dual-reading it re-reads the
pre-hardening collision bucket. The collision is closed for **NEW (hardened) WRITES only**;
legacy data keeps its pre-#661 collision semantics (the same accepted behavior under the
single-owner v1 threat model, where userIds are not adversarially controlled). Dual-read is
strictly **≥** the pre-#661 state (it can never lose data) but does **NOT** retroactively
split already-colliding legacy rows — structurally impossible (one shared bucket cannot be
disambiguated). Asserted in a `HONEST:` test in both TS and Rust.

### What remains
- **Operator gate to flip `FRIDAY_NS_HARDENING_ENABLED=1` in prod** (the deploy-GO; not done here).
- **Write-path hardening is partial under flag-on (DARK-remaining):** the WRITE paths
  (extraction-service + agent `memory_store`) source the namespace from the **cached**
  `session.memoryNamespace`, so flipping the flag ON makes new writes HARDENED only for
  brand-new sessions (`createSession` re-derives); pre-existing sessions keep writing to
  their cached legacy namespace until `alignSessionContext` recomputes. **Data-safe either
  way** (dual-read finds legacy), but collision-closure (c) applies only where writes are
  actually hardened. Wiring the write paths to re-derive under flag-on is intentionally
  deferred (it touches the broad live write path; surfaced as an operator question rather
  than blind-changed).
- **Rust live recall-principal wiring (DARK-remaining):** the Rust recall principal is the
  configured `--owner` allowlist entry (`RunPolicy::principal_id`), NOT a session-derived
  namespace, so `recall_confirmed_multi` + the candidate builder are the GATED MECHANISM;
  re-wiring the live recall principal to consult the candidate list is not done here.

### Tests (honest counts, all run locally)
- **TS:** `friday-session-memory-namespace.test.ts` 22 pass (flag-off byte-identical,
  flag-on hardened, dual-read `[hardened,legacy]`, dedup-collapse, collision-closed,
  HONEST legacy-collision). `friday-agent-memory-tools.test.ts` 44 pass (incl. dual-read
  THROUGH the live consumer: legacy-bucket memory recalled via semantic AND lexical
  fallback; flag-off single query; dedup-by-id; older-dep fallback).
  `friday-session-service.test.ts` 99 pass (incl. flag-off→on: a session persisted legacy
  is still recalled via the legacy candidate). 215 pass across dependent suites
  (subagent-registry, extraction-service, api-runtime). `tsc --noEmit` clean; `npm run lint`
  0 errors; `check:secret-patterns` clean.
- **Rust:** `session_namespace` 26 lib unit tests pass; `friday-hub` lib 520/520;
  `memory_persist` 18 (incl. `recall_confirmed_multi` union/dedup + single-principal
  isolation). `cargo clippy -p friday-hub -p friday-storage --all-targets -D warnings`
  clean. `cargo fmt` (pinned 1.95.0) applied + `--check` clean.

DARK + DEPLOY-GO-gated. No live wiring/deploy. Do NOT merge.

---

### Update — additional verification
- **Rollback (on→off) data-safety test added.** A `getOrCreateSession` under flag-on persists a HARDENED `memory_namespace`; after flipping the flag OFF, the re-derived candidates are `[legacy]` — and the defensive cache tail appends the persisted hardened namespace, so hardened-written rows are STILL recalled. This pins the REVERSE direction of the guarantee (no loss on rollback). 100 session-service tests pass.
- **Write-persist nuance (clarifies the partial-write concern):** `getOrCreateSession` (the live runtime path) PERSISTS the resolved namespace; `createSession` does not (lazy re-derivation). So under flag-on, brand-new sessions via `getOrCreateSession` persist hardened; the dual-read + defensive tail keep every other path data-safe.
- **Full `vitest run --project default`:** 12,872 passed, 159 skipped, **1 pre-existing unrelated failure** (`friday-secret-crypto.test.ts` — `FRIDAY_MASTER_KEY_SOURCE=keychain`, env-dependent; fails identically on `main` without this diff). No failure is attributable to this change.

### DIFF
--- a/rust-core/crates/friday-hub/src/session_namespace.rs
+++ b/rust-core/crates/friday-hub/src/session_namespace.rs
@@ -7,11 +7,26 @@
 //! user's memory is isolated PER (account, channel) instead of being shared across every
 //! channel for the same user.
 //!
-//! Format (byte-identical to the TS — including the Unicode-aware lowercasing; see the
-//! normalization note on the dot-join residual that is SHARED with the TS oracle):
+//! Format (byte-identical to the TS — including the Unicode-aware lowercasing; the dot-join
+//! collision hardening is FLAG-GATED + DEFAULT-OFF + DUAL-READ — see the normalization note
+//! on [`is_kept`] and the dual-read note on
+//! [`resolve_session_memory_namespace_candidates`]):
 //! ```text
 //! tenant.<accountId>.channel.<channel>.user.<normalizedUserId>.shared
 //! ```
+//!
+//! ## Collision hardening (F5.5): FLAG-GATED, DEFAULT-OFF, DUAL-READ
+//! Dropping `.` from the segment keep-set closes the dot-join cross-tuple collision, but a
+//! naive deploy of that change RE-SCOPES (orphans) any existing memory whose segment
+//! legitimately contained a `.` (an email-shaped userId, a dotted account/channel) — a
+//! one-way data downgrade. So the hardening is gated behind the DEFAULT-OFF env flag
+//! [`NS_HARDENING_ENV_FLAG`]:
+//! - OFF (default): the keep-set KEEPS `.` exactly as before — every namespace is
+//!   BYTE-IDENTICAL to the legacy derivation. Zero re-scope on deploy.
+//! - ON: the WRITE/primary namespace is hardened, AND the READ path dual-reads BOTH the
+//!   hardened and legacy namespaces ([`resolve_session_memory_namespace_candidates`]), so
+//!   nothing written under the legacy namespace is lost. There is NO destructive re-key.
+//! PARITY: the TS half reads the SAME flag name under the SAME `"1"` semantics.
 //!
 //! ## How this binds to the Rust recall axis
 //! The Rust memory recall axis is a single `principal_id` string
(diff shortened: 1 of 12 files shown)


thesongzhu:
QUEUED — DO NOT MERGE. 2/3 review BLOCK: flag-OFF LIVE recall regression (when persisted session.memoryNamespace differs from re-derived legacy namespace, e.g. post-rollback). ns-hardening is now 2-for-2 problematic on the LIVE recall path (orig #661=downgrade, this=flag-off regression). It's non-GO-critical (F5.5). Recommend operator decide: one more careful fix (+ the adversarial flag-off test the reviewers specified) OR drop the re-scope (additive-only). Not blind-iterating.

PR #675, opened 2026-06-12
feat(ui-read-seam): S-R0 shared sealed-WS substrate + S-R1 dark read-projection server (workbench) [DARK, default-off]

thesongzhu:
## What

The FIRST dark slice of the v1 UI **Rust direct-read seam** (per the locked FILE55 decision: the v1 SwiftUI UI direct-connects to the Rust hub over the DARK sealed-WS, NOT through a TS hub route). Operator-locked transport = **2b** (a SEPARATE sealed-WS read server) on a **2c** shared substrate. Design of record: `UI_RUST_DIRECT_READ_SEAM_DESIGN_20260611.md`.

**ALL DARK, default-off.** No new LaunchAgent, no production caller. The live WRITE path is byte-identical. This moves **zero UI needle** until D-PR1 consumes it AND the slice-6 production flip lands (operator-gated). NOT v1 GO.

## S-R0 — shared sealed-WS crypto/auth substrate (`friday_hub::sealed_ws`)

Factored the common transport/auth **primitives** (NOT the serve loop — the write loop is saturated with write-only concerns; genericizing it would be the most likely way to break byte-identical) into one shared module that BOTH servers use, so crypto/auth parity is guaranteed by construction:

- `establish_session` — the cleartext preamble (peer pubkey in → **S-F peer-pubkey-allowlist gate FIRST** → low-order check → server pubkey out → fresh CSPRNG nonce out → WS upgrade) + ECDH session key.
- the peer-allowlist boot chain (`load_peer_allowlist` / `peer_is_allowlisted` / `enforce_single_peer` / `PeerAllowlistError`), the `LOW_ORDER_X25519_POINTS` table + `is_low_order_x25519`, the sealed-proof codec (`encode`/`decode_sealed_proof`), `SESSION_NONCE_LEN`.
- `AuthedPrincipal::authenticate_forwarded` (in `hub_server.rs`) generalized: `auth_aad` / `ForwardedAuth` now bind an **opaque `bound_context: &[u8]`** instead of a `run_id`-typed field, so a read (which has no run) binds its own per-request id and keeps the IDENTICAL nonce-bound + owner-allowlisted + possession-of-session guarantee.

### Byte-identical-for-writes argument (deploy safety)

A read has no run, so the AAD binding was generalized from `run_id: &str` to `bound_context: &[u8]`. **The write path passes `run_id.as_bytes()`** — exactly the bytes the prior `auth_aad` took internally — so the emitted auth AAD is **byte-identical** for writes (a 4-byte BE length prefix over the same bytes). The write bin re-expresses its prior behavior through the shared fns with the SAME constant bytes and the SAME preamble framing. Proof:
- **frozen-AAD KAT** (`auth_aad_write_path_is_byte_unchanged`) pins the exact `Vec<u8>` for a known `(SESSION_AAD, principal, run_id)` — a literal, not an old-fn==new-fn comparison.
- the write bin's **full 32-test real-handshake suite stays green** (dispatch round-trip + body-sealed-never-in-refs, forged peer, captured-proof replay, low-order, wrong-key, non-allowlisted principal/peer, msg_id replay) — stronger evidence than the KAT alone.

## S-R1 — separate DARK sealed-WS read server (`hub_read_projection_server`)

New bin on the S-R0 substrate that:
- opens the hub DB **READ-ONLY** (`Db::open_hub_readonly`), builds **NO `HubRuntime` and NO `DeepSeekClient`**, never calls `from_env` — **credential/capability isolation by construction**: no model-call path, no provider credential, can only read. A read client cannot spend quota or write. The absence of the runtime IS the no-model-path guarantee.
- accepts `WorkbenchProjectionRequest` (named projection + params + forwarded principal + `auth_proof` bound to a per-request `request_id`), authenticates against the sealed session via the SAME owner-auth chain a write uses,
- runs the extracted shared `friday_hub::workbench_projection::project_workbench` fn (refs-only; the **forbidden-output guard runs INSIDE** it, so the bin AND the server inherit refs-only),
- returns an **OWNER-SEALED, REFS-ONLY** `WorkbenchProjectionSnapshot` (the projection JSON is double-sealed under the session key; refs like ledger/result/activity ids + counts + labels, never an inline body; truth labels as-is, never upgraded). A `503`/missing/non-canonical state is surfaced as a **typed `Error`**, never a fabricated success.
- **owner-scoped**: released ONLY to the bound AUTHENTICATED principal, never a client-asserted id; a forged/empty/non-allowlisted principal ⇒ no snapshot, end the session.

The `mission_workbench_projection` CLI bin is now a **thin wrapper** over the shared lib fn (one implementation, no drift).

> **Domain separation (for the future Swift client):** the read server's `SESSION_AAD` / `AUTH_CHALLENGE` are `:ui-read-seam:…:s-r1:` — DISTINCT from the write constants. A read client MUST seal with the read constants, not the write ones.

## Which KATs prove what

| KAT | proves |
| --- | --- |
| `hub_read_projection_server::workbench_projection_round_trips_through_the_read_server` | full seam end-to-end: handshake → owner auth → owner-sealed refs-only snapshot; the owner-opened body is the canonical refs-only projection with no secret markers |
| `hub_read_projection_server::read_server_rejects_a_mismatched_principal` | owner-scoping: a well-formed but non-allowlisted forwarded principal gets NO snapshot, the session ends (processed==0) |
| `hub_server::…::auth_aad_write_path_is_byte_unchanged` | the write-path auth AAD is byte-identical after the `bound_context` generalization (frozen literal) |
| `sealed_ws::tests::*` | substrate units: peer-allowlist fail-closed (missing/invalid/multi-peer), low-order rejection (incl. high-bit variant), sealed-proof codec round-trip |
| write bin's existing 32-test suite (unchanged) | byte-identical write behavior through the refactor |

## DEFERRED (explicit acceptance criteria — NOT faked)

- **S-R2..S-R5** — the other UI read projections (run-readback / providers-doctor / diagnostics / workflow-catalog). Three already have library fns; run-readback needs the same small extraction.
- **Cross-owner / per-conversation `owner_principal` isolation.** This slice is **single-owner v1**: the owner-allowlist ceiling is the SAME single-configured-owner model the write path uses today; the read server returns the first active mission without matching `mission`/conversation `owner_principal` to the caller. Safe ONLY because v1 has one owner. Per-conversation owner-match (so peer A cannot read peer B's mission even with a valid session) is tied to the write path's own unbuilt FIX-Q2 / FIX-Q3b multi-principal bindings — deferred, not silently skipped.
- **G3 — the SwiftUI sealed client.** Net-new; the crypto-parity surface (X25519 + HKDF-SHA256 32-zero-salt + XChaCha20-Poly1305 24-byte nonce + `[nonce_len][nonce][ciphertext]` framing) is the hard part. Operator/device-gated.
- **G4 — UI peer-pubkey enrollment.** The UI's X25519 pubkey must enroll in the SecureStore `PEER_PUBKEY_ALLOWLIST_ID` allowlist (the provisioning/device gate). Single-peer is a server invariant until the multi-principal bindings land.
- **slice-6 production flip (G2, the FREEZE tripwire)** — provision the read server's LaunchAgent + enroll the UI peer + flip the prod flag. Operator-gated; NOT this slice.

## localChecks

- `cargo +1.95.0 fmt --all` — clean
- `cargo +1.95.0 clippy --workspace --all-targets` — clean (0 warnings)
- `cargo +1.95.0 build` — clean
- `cargo +1.95.0 test -p friday-hub -p friday-protocol -p friday-ffi -p friday-crypto` — all green; write bin 32/0, read bin 2/2, sealed_ws 6/6, frozen-AAD KAT pass
- `cargo +1.95.0 test --workspace` — see PR thread for the one `friday-anthropic` real-socket test result (untouched crate; passes 18/18 per-crate)


thesongzhu:
### localChecks — `cargo +1.95.0 test --workspace` characterization (honest)

The workspace run shows intermittent failures confined to **`friday-anthropic`** real-transport tests (e.g. `real_transport_sends_x_api_key_…`, `real_transport_maps_rate_limit_…`) — panics with `BadResponse("invalid JSON: … Invalid argument (os error 22)")` and inside `ureq-2.12.1/src/response.rs`. This is a **pre-existing intermittent flake in an UNTOUCHED crate**, not a regression from this PR. Evidence:

1. **`git diff origin/main -- rust-core/crates/friday-anthropic/` is EMPTY** — the crate is byte-identical to main; nothing in this PR can have caused it.
2. **The failing set VARIES between runs** (one workspace run failed 1 test; the next failed 2, including a *different* test) — the signature of an intermittent real-socket/HTTP flake, not a deterministic break.
3. **Running `cargo test -p friday-anthropic real_transport` alone, repeatedly, is itself intermittent**: 5 consecutive runs → FAILED, ok, ok, ok, ok. Same source, varying outcome. The tests bind localhost sockets and read HTTP responses; under concurrent socket pressure they intermittently hit OS error 22.
4. `cargo test -p friday-anthropic` (full crate) passes **18/18** in isolation.

**This PR's own crates are deterministically green:** `cargo test -p friday-hub -p friday-protocol -p friday-ffi -p friday-crypto` — write bin **32/0**, read bin **2/2**, `sealed_ws` **6/6**, frozen-AAD KAT pass; `clippy --workspace --all-targets` clean; `fmt` clean; `build` clean.

PR #736, opened 2026-06-14
feat(f11 tts-1): friday-fs create_dir_all_within_root root-contained mkdir primitive

thesongzhu:
## F11 Wave-2 · TTS-1 (friday-fs)

Net-new pure-Rust primitive. `friday-fs` today has **no mkdir**: `write_file_within_root` / `open_write_within_root` both require `parent_must_exist=true` and explicitly do not create intermediate directories. The TTS artifact path is nested (`.friday/artifacts/tts/tts-<ts>.<fmt>`), so the later TTS-5 executor cannot create it — this PR adds the missing primitive.

### What it does
`pub fn create_dir_all_within_root(root: &Path, candidate: &str) -> Result<(), FsError>` — a root-contained, symlink-escape-safe `mkdir -p` that **reuses the same `resolve_within_root` lexical + canonicalized-root + realpath-ancestor containment gate** as every other primitive in the crate.

- **Top-down per-component walk** (not `std::fs::create_dir_all`): builds each prefix `a`, `a/b`, `a/b/c`, runs the full `resolve_within_root` on it, then `create_dir`s exactly one new level. By the time level N is resolved every ancestor already exists as a verified real dir, so the realpath-ancestor check fires for every component and the kernel never resolves a parent **through a symlink** — the exact escape the naive `create_dir_all` shortcut would silently follow.
- **Idempotent** on an existing real directory.
- **Refuses** a symlink at any component (`FsError::Symlink`, even an in-root-pointing one — same don't-follow posture as delete/move/write) and a file-at-path (`FsError::NotADirectory`).
- **Lexical escapes** (`..` / absolute / `.` / empty) reject up front as `FsError::Lexical` with no directory created as a side effect.
- Ancestor-component TOCTOU is documented as the same accepted residual the rest of the crate carries.

### Dark discipline
Inert primitive: **no flag, no external dep, no hub / ToolExecutor / registry / capability touch** (the first caller is TTS-5 in a later wave). Touches `rust-core/crates/friday-fs/src/lib.rs` ONLY.

### Tests (7 in-lib `#[cfg(test)]`, real on-disk dirs + real symlinks)
nested-create · idempotent + fill-missing-tail · lexical reject (absolute / `..` / nested-`..` / empty, asserting no partial side effect) · file-at-intermediate (`NotADirectory`, file intact) · **the load-bearing ancestor-symlink-escape** (negative assertion: the outside dir gains no `b`/`c`) · final-component symlink · missing-root `NotFound`.

### Gate (from rust-core, all green)
- `cargo fmt --all -- --check` ✓
- `cargo clippy -p friday-fs --all-targets -- -D warnings` ✓ (no warnings)
- `cargo build -p friday-fs` ✓
- `cargo test -p friday-fs` ✓ — all 78 friday-fs tests pass (verified single-threaded + repeated runs); my 7 new tests pass every run
- `cargo build --workspace` ✓ (friday-fs integrates with friday-hub which consumes it)

> Note: the pre-existing `run_command_setsid_escaped_grandchild_does_not_pin_thread` test (unmodified by this PR; not in the friday-fs primitives I touched) is a known scheduler-jitter flake — its own doc comment notes loaded-CI sensitivity and it passes reliably single-threaded and in isolation. This PR makes no change to `run_command`.

### DIFF
--- a/rust-core/crates/friday-fs/src/lib.rs
+++ b/rust-core/crates/friday-fs/src/lib.rs
@@ -1342,6 +1342,125 @@ pub fn run_command_in_root_with_timeout(
     })
 }
 
+/// Recursively create a contained directory and any missing intermediate directories
+/// (`mkdir -p`), root-contained and symlink-escape-safe. **Idempotent**: an already-existing
+/// contained directory is `Ok(())`.
+///
+/// # Why this exists
+/// friday-fs has no `mkdir`: [`open_write_within_root`] / [`write_file_within_root`] both
+/// require the parent directory to already exist and explicitly do NOT create intermediates
+/// ("we do not `mkdir` intermediate directories"). A nested artifact path such as
+/// `.friday/artifacts/tts/tts-<ts>.<fmt>` therefore cannot be written by the existing
+/// primitives until its directory chain is created. This primitive creates that chain while
+/// preserving the SAME containment guarantees as every other primitive in this crate.
+///
+/// # Containment (reuses the shared [`resolve_within_root`] gate, top-down per component)
+/// We do NOT hand the full nested path to `std::fs::create_dir_all`: that lets the **kernel**
+/// resolve the parent portion, which would FOLLOW a pre-existing ancestor symlink and create
+/// directories *outside* the root (the exact escape this primitive must refuse). Instead we
+/// walk the candidate's components **top-down**, and for each prefix `a`, `a/b`, `a/b/c`, …:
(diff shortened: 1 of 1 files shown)


thesongzhu:
Closing per the 2026-06-13 strategic decision (Desktop FRIDAY_战略方向决策_20260613.md): F11 (vision/OCR/TTS/browser/PDF) is a Rust re-implementation of capabilities the TS product already runs live, and is post-v1 — not on the Phase-1 operating-layer spine. Work is banked; reopen in Phase-2 when we reach per-capability Rust parity.

PR #742, opened 2026-06-14
fix(rust-agent-run): surface honest non-Finished terminal status instead of misleading readback_not_found 503

thesongzhu:
## Root cause

In `composeRustReadOnlyAgentRun` (`src/api/runtime/friday-api-runtime.ts`), the owner-gated answer readback's `outcome !== "delivered"` branch ALWAYS threw `failClosed()` → 503 `readback_not_found`.

But a run whose Rust loop terminates **NON-Finished** (LoopStatus Blocked / Errored / Bounded, or fail-closed-to-no-answer) **deliberately skips the Rust-side persist guard** (`if outcome.status == Finished { persist_run_result }` in the hub runtime). No `run_result` row is ever written, so the owner-gated body readback **legitimately** returns `not_found` — there was never an answer to store. Throwing the 503 for that case **misreports an honest terminal settle as a transport/storage failure**. This hit the hourly self-probe and S6.

The dispatch loop-status label (`wsResult.status`) is already in scope at the readback decision point.

## The fix — Case-A / Case-B split (no gate-weakening)

At the `outcome !== "delivered"` branch we now carve out **only** the honest non-Finished terminal case with a **strict allowlist**, and otherwise keep today's 503 exactly:

- **Case B — honest non-Finished terminal (NO 503):** `outcome === "not_found"` **AND** `wsResult.status ∈ { no_answer_safe_failure, no_answer, blocked, bounded, errored }`. The wire collapses Blocked/Errored/Bounded/no-answer into one label (`AuthedAnswer::NoAnswer`'s `proof_refs_json` carries no `status` key → the agent-run server falls back to the `outcome`, so compose actually sees `"no_answer_safe_failure"` for all non-Finished/no-body terminals; the raw loop-status tokens are admitted too). We project an HONEST refs-only continuity row via the **same write-transaction + projector pattern the Paused branch uses**, mapped to `loopStatus:"Errored"` → the projector's terminal **"failed"** status (never `"Paused"`/`"cancelled"` — a no-answer terminal is not resumable). Refs-only: empty body ref (empty-string sha sentinel + `len 0`), empty response — **never a fabricated answer body**. Owner is stamped so the row is owner-scoped for the read routes. We deliberately do **not** stamp the apiRequest idempotency descriptor (a failed/no-answer run must not replay on a retry).

- **Case A — genuine error (KEEP 503):** every other non-delivered case falls through to the unchanged `logFailClosed()` + `throw failClosed()`:
  - `wsResult.status === "finished"` + `not_found` → a finished run **should** have a row, so a missing row is a real failure → **still 503**.
  - `outcome === "denied"` (an ownership-gate refusal) → **still 503** even with a non-Finished status.
  - any error status (`storage_failed` / `auth_failed` / `operator_vk_unprovisioned`) → not in the allowlist → **still 503**.

### No-degrade argument

The carve-out is a strict allowlist requiring **both** conjuncts; everything else defaults to today's fail-closed 503. A finished run with a missing row still 503s; `denied` still 503s; any storage/auth error still 503s. We never fabricate an answer body. The read-only delivered path and the Paused path are untouched.

## KAT (colocated, `friday-api-runtime-rust-route-compose.test.ts`)

- `(nf-a)` `no_answer_safe_failure` (the value the live wire emits) + `not_found` → **NO 503**; status `"failed"`, one continuity row with `loopStatus "Errored"`, empty body ref `len=0`, no body, owner stamped.
- `(nf-b)` `blocked` + `not_found` → same (the task's named case).
- `(nf-c)` Case A: `finished` + `not_found` → **STILL 503**, no row.
- `(nf-d)` no-weakening: `storage_failed` + `not_found` → **STILL 503**, no row.
- `(nf-e)` no-weakening: `denied` (with a non-Finished status) → **STILL 503**, no row.

## Constraints / hygiene

- **TS-only**: touches only the compose source + its colocated test. No Rust, no other hot files.
- Born-current on `origin/main` `70b458ca`.
- Local green: `npm run lint` (0 errors), `npm run typecheck` (0 errors), `npm test` for this file (**34/34 pass**). Both `secrets`-job gates pass (detect-secrets-hook + provider-key-pattern denylist); the empty-string SHA constant is a public well-known hash with an allowlist marker.
- The honest-terminal projection is extracted to a module-level helper so the compose function stays under the per-function line limit.

## Deploy note

This changes **source only**. Making it live on prod needs a `dist` rebuild + hub restart — a **separate operator-gated deploy**, not attempted here.

### DIFF
--- a/src/api/runtime/friday-api-runtime.ts
+++ b/src/api/runtime/friday-api-runtime.ts
@@ -993,6 +993,42 @@ export const RUST_ROUTE_READ_TOOL_ALLOWLIST = ["read_file", "list_dir", "stat_fi
 const RUST_ROUTE_DEEPSEEK_PROVIDER_ID = "deepseek";
 const RUST_ROUTE_DEEPSEEK_FLASH_MODEL = "deepseek-v4-flash";
 
+// (honest-non-finished) The well-known SHA-256 of the EMPTY byte string. A non-Finished terminal
+// run produced NO answer body, but the continuity-projector receipt requires a `finalMessageSha256`
+// string — we stamp this empty-body sentinel (with `finalMessageLen: 0`) so the body REF is
+// truthfully "zero bytes", NEVER an answer fingerprint. Refs-only — this is not a body.
+const EMPTY_BODY_SHA256 =
+  "e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855"; // pragma: allowlist secret
+
+// (honest-non-finished) The CLOSED allow-list of dispatch loop-status labels that mark a
+// NON-deliverable terminal run (the run ended without an answer to persist). The wire COLLAPSES
+// Blocked/Errored/Bounded/no-answer into ONE label: `AuthedAnswer::NoAnswer`'s refs carry no
+// `status` key, so the agent-run server falls back to the `outcome` → compose actually sees
+// `"no_answer_safe_failure"` for ALL non-Finished/no-body terminals (the raw loop-status tokens are
+// admitted too, for a server that surfaces them directly). EVERY OTHER status — notably `"finished"`
+// (a finished run SHOULD have a row) and any error status (`storage_failed`/`auth_failed`/
+// `operator_vk_unprovisioned`) — is NOT here, so it keeps today's fail-closed 503 (no gate-weakening).
+const RUST_ROUTE_NON_FINISHED_TERMINAL_STATUSES = new Set<string>([
+  "no_answer_safe_failure",
+  "no_answer",
+  "blocked",
+  "bounded",
+  "errored",
+]);
+
+// (honest-non-finished) Map a NON-Finished terminal wire status to the projector's `loopStatus`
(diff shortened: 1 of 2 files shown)


thesongzhu:
Post-review refinement (same two files, still TS-only, all green): the projected non-Finished `loopStatus` is now FAITHFUL where the wire allows it — a raw loop-status token (`blocked`/`bounded`/`errored`) keeps its capitalized form (Blocked/Bounded/Errored), while the opaque collapsed labels the live wire actually emits (`no_answer_safe_failure`/`no_answer`) fall back to `Errored`. All of these still map to the SAME product-visible terminal status `"failed"` (mapLoopStatusToTsStatus) — the distinction is metadata fidelity only, never a different outcome, and never the resumable `Paused`/`cancelled`. KAT `(nf-b)` updated to assert the faithful `Blocked` label; `(nf-a)` keeps `Errored` for the opaque live-wire value. Full `test/unit/api/runtime/` suite green (196/196).

PR #744, opened 2026-06-14
feat(mission-spine): organic POST routes (intake/lifecycle/work-item-status) via sealed-WS, flag-gated dark

thesongzhu:
## What & why (Lane B)

The **organic caller** for the mission-spine. The keystone #741 wired the Rust `hub_agent_run_server` sealed-WS **dispatch arms** for Mission intake / Mission lifecycle / WorkItem status (gated behind server flags, default-OFF), but nothing in the product reaches them. This PR adds the TS half — sealed-WS client methods + flag-gated POST routes — so the UI/API can actually **create + transition** Missions/WorkItems and **Loop1 can close live** (once the operator flips the server flags + deploys dist).

## The 3 routes (mirror the GET `/v1/mission-spine/workbench` contract)
| Method | Path | operationId | Client method → wire |
|---|---|---|---|
| POST | `/v1/mission-spine/intake` | `mission.spine.intake.create` | `intakeMission` → `MissionIntakeRequest` |
| POST | `/v1/mission-spine/:missionId/lifecycle` | `mission.spine.lifecycle.transition` | `transitionMission` → `MissionLifecycleRequest` |
| POST | `/v1/mission-spine/work-items/:workItemId/status` | `mission.spine.workitem.status.transition` | `transitionWorkItem` → `WorkItemStatusRequest` |

**Handler guard order (per route):** (1) dispatch-disabled (flag-OFF) → **503 honest-unavailable FIRST**, uniform regardless of caller; (2) **bound-principal** (`assertBoundPrincipalForOperation`) → refuse the synthetic public principal (401); (3) body validation → typed **400**; (4) seal + dispatch over the sealed-WS client; refs-only result passthrough. `proof_receipt` is a passthrough on the work-item route (Rust enforces proof-on-completion).

## Client methods (`friday-rust-hub-agent-run-ws-sealed-client.ts`)
- `intakeMission` / `transitionMission` / `transitionWorkItem` mirror `resumeWithApproval`'s handshake (RAW connect → length-prefixed preamble → X25519+HKDF session key → manual RFC6455 upgrade → seal+send → await the FIRST matching `*Result` → settle) via one shared `runMissionRoundTrip` helper, reusing the existing module-level helpers. **`dispatchRun` is untouched** (its `AgentRunRequest` byte-identity is preserved).
- Pure, exported, **socket-free-testable** builders/parsers that map to the **EXACT** `friday-protocol` wire shapes. Option fields use conditional-spread (absent ⇒ key omitted), mirroring `buildConstraintsWire`. These carry **no `auth_proof`/`forwarded_principal`** — the sealed session **is** the channel auth (single-peer/single-owner SERVER invariant), as the keystone arms document.
- **Refs-only + fail-closed:** ANY non-matching inbound — **including the server's `Error` envelope** (illegal hop / missing entity / proofless completion) — is a typed fail-closed (503), never a partial. (Granular 4xx mapping of a server-side domain rejection is deferred; today it's a clean fail-closed 503.)

## The flag & dark-safety (flag-OFF byte-identical / honest-unavailable)
- The route flag lives as `deps.dispatch` on `FridayMissionSpineRoutesDeps`, **default `null`**. `null` ⇒ each POST route is honest-unavailable (503), the **same pattern** as the read route's unavailable path. The base/contract runtime (no `deps.missionSpine.dispatch`) keeps it `null` automatically, so **adding these routes is byte-identical to today for existing traffic** — no live-by-default mutating route, no runtime DI change needed.
- **LIVE closure additionally needs the SERVER flags** (operator-gated flip) + a **dist deploy**. Note the two-flag nuance from #741: intake rides **`FRIDAY_MISSION_INTAKE`**, while lifecycle + work-item ride **`FRIDAY_MISSION_SPINE_DISPATCH`** (the task statement collapsed these into one).

## Gate family (no gate-weakening)
The 3 public mutating routes are classified **`bound_principal`** (NOT dropped into the `unclassified` debt bucket — that would be a new ungated mutating surface). Each handler calls `assertBoundPrincipalForOperation` before any dispatch; 3 new members were added to the `FridayPublicMutationOperation` union and the WP-001 contract `BOUND_PRINCIPAL` set. The contract snapshot was regenerated and the diff is **verifiably additive**: 3 new entries + index renumber + route count 420→423; `bound_principal` 59→62, `total_public_mutating` 321→324; **no existing route's auth/path/method changed**.

## KATs
- **flag-OFF** → routes honest-unavailable 503, **no client call** (incl. for a `null` principal — 503 fires before the principal check).
- **bound-principal** → a `null`/synthetic public principal is refused with 401, **no dispatch**.
- **flag-ON** → each route builds the correct wire + calls the (mocked) client + returns the refs-only result; path id (`missionId`/`workItemId`) is taken from the path; optional fields omitted when absent.
- **invalid body / missing path id / proofless `completed_with_proof`** → typed 400, fail-closed (no dispatch).
- builder optional-omission (snake_case, bool only when true) + parser **fail-closed on missing/ill-typed required refs** (all `*Result` fields verified field-for-field against the origin/main wire structs, incl. the required `reason` on `WorkItemStatusResult`).

## Local gates
- `npm run lint` — **0 errors** (warnings only, matching existing codebase patterns).
- `npm run typecheck` — clean (src + contracts).
- `npx vitest run` mission-spine + routes — **35 new tests** green; full mission-spine suite (214) green.
- `npm run test:contracts` / `test:contracts:routes` — green (snapshot updated, additive).
- `npm run check:secret-patterns` — clean.

## Scope
TS-only. Touches the mission-spine routes file, the sealed-WS client, the public-mutation-op union (1 line family), and the WP-001 contract test set. **No Rust, no unrelated hot files, no runtime DI change.** Born current on the latest `origin/main` at branch time. Do **not** merge — this lands DARK; the product needle moves only on the operator's server-flag flip + dist deploy.

### DIFF
--- a/src/api/http/routes/friday-mission-spine-routes.ts
+++ b/src/api/http/routes/friday-mission-spine-routes.ts
@@ -1,10 +1,19 @@
 import { FridayDomainError } from "#errors";
+import { assertBoundPrincipalForOperation } from "../../../security/friday-owner-session-channel-capability.js";
 import type { FridayRouteDefinition } from "../../model/friday-api-common.types.js";
 import type {
   FridayMissionSpineLifecycleState,
   FridayMissionSpineWorkbenchResponse,
   FridayMissionSpineWorkbenchSnapshot,
 } from "../../model/friday-api-mission-spine.types.js";
+import type {
+  FridayRustHubMissionIntakeRequest,
+  FridayRustHubMissionIntakeResult,
+  FridayRustHubMissionLifecycleRequest,
+  FridayRustHubMissionLifecycleResult,
+  FridayRustHubWorkItemStatusRequest,
+  FridayRustHubWorkItemStatusResult,
+} from "../../mission-spine/friday-rust-hub-agent-run-ws-sealed-client.js";
 
 export interface FridayMissionSpineWorkbenchProjectionInput {
   missionId?: string;
@@ -17,9 +26,48 @@ export interface FridayMissionSpineWorkbenchProjectionService {
   getSnapshot(input: FridayMissionSpineWorkbenchProjectionInput): Promise<FridayMissionSpineWorkbenchSnapshot>;
 }
 
+/**
+ * (Lane B) The ORGANIC mutation seam: a thin service that drives the (flag-gated) Rust sealed-WS
+ * dispatch arms for the three Hub-owned mission-spine mutations. The route handlers validate the
+ * body, then hand a typed request to this service, which seals + sends over the sealed-WS client
+ * and returns the refs-only result. When this service is `null` (the DEFAULT — the route flag is
+ * off / no client wired), the POST routes return honest-unavailable (503), byte-identical to the
+ * read route's unavailable path. PROVIDING it is what makes the routes LIVE — and live closure
+ * ALSO needs the SERVER flags (`FRIDAY_MISSION_INTAKE` for intake, `FRIDAY_MISSION_SPINE_DISPATCH`
+ * for lifecycle/work-item), which are a separate operator-gated flip.
+ */
+export interface FridayMissionSpineDispatchService {
+  intakeMission(request: FridayRustHubMissionIntakeRequest): Promise<FridayRustHubMissionIntakeResult>;
+  transitionMission(
+    request: FridayRustHubMissionLifecycleRequest,
+  ): Promise<FridayRustHubMissionLifecycleResult>;
+  transitionWorkItem(
+    request: FridayRustHubWorkItemStatusRequest,
+  ): Promise<FridayRustHubWorkItemStatusResult>;
+}
+
 export interface FridayMissionSpineRoutesDeps {
   readonly workbench: FridayMissionSpineWorkbenchProjectionService | null;
+  /**
+   * (Lane B) The organic mutation dispatcher, DEFAULT-OFF (`null`). `null` ⇒ the three POST routes
+   * are honest-unavailable (503); a real adapter ⇒ they seal + dispatch over the sealed-WS client.
+   * Adding the routes with this `null` is byte-identical to today for existing traffic.
+   */
+  readonly dispatch?: FridayMissionSpineDispatchService | null;
+  /** (Lane B) Optional reason for the dispatch-unavailable 503; falls back to a default message. */
+  readonly dispatchDisabledReason?: string | null;
   readonly disabledReason: string | null;
+}
+
+/** (Lane B) Response envelope for each organic mutation route — refs-only result passthrough. */
+export interface FridayMissionSpineIntakeResponse {
+  readonly result: FridayRustHubMissionIntakeResult;
+}
+export interface FridayMissionSpineLifecycleResponse {
+  readonly result: FridayRustHubMissionLifecycleResult;
+}
+export interface FridayMissionSpineWorkItemStatusResponse {
+  readonly result: FridayRustHubWorkItemStatusResult;
 }
 
 const DEFAULT_DISABLED_MESSAGE =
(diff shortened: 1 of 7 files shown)


thesongzhu:
## Fix: HOLE-1 — Mission wire fields nested under `request`/`result` (commit 871d2dd9)

**Root cause (confirmed):** `friday-protocol`'s `Message` enum is `#[serde(tag = "kind")]` (internally tagged), and the three Mission variants are **single-field wrappers** — `MissionIntakeRequest { request: MissionIntakeRequestWire }`, `MissionIntakeResult { result: MissionIntakeResultWire }`, and likewise lifecycle + work-item. serde therefore **nests** the inner wire fields under a `request`/`result` key, not flat under `message`. This PR originally built/parsed them flat, so:
- **outbound** → the server's `Envelope::decode` (serde) failed on the missing `request` key → session dropped → client 503;
- **inbound** → `parse*Result` read `fields.X` instead of `fields.result.X` → `undefined` → required-ref check failed → 503.

Every intake/lifecycle/work-item call 503'd; Loop1 could not close.

**Fix (client + wire test only, TS-only, wire-shape only):**
1. The 3 `build*Envelope` fns now wrap the field object under `request`: `message: { kind, request: { …fields… } }`.
2. The 3 `parse*Result` fns now unwrap `fields.result` first, via a shared `unwrapResult` helper that **fails closed without throwing** (parse runs outside the inbound try/catch in `runMissionRoundTrip`; a missing/ill-typed `result` → `undefined` → clean 503, never a TypeError).
3. The 6 build/parse test expectations were updated to the nested shape, plus added fail-closed tests for the flat (pre-fix) shape.

**Golden cross-check (the real fix for the mock-green trap):**
I ran the Rust `Envelope::encode` round-trip in `rust-core/crates/friday-protocol/src/lib.rs` (via a throwaway print test, since removed — the protocol crate is unchanged in this PR) and captured the **exact bytes** it emits for all 6 messages. The wire test now pins:
- (a) each `build*Envelope`'s `message.request` **deep-equals** the Rust-emitted `request` sub-object, and
- (b) each `parse*Result` **consumes** the Rust-emitted `result` JSON.

Sub-object deep-equal (not full-envelope byte-equal) is deliberate: `sent_at` (=`Date.now()`), `msg_id` (per-entity, not `"m1"`), and `schema_version` (TS 12 vs current Rust 13) are orthogonal to the nesting bug. The captured fixtures break-on-drift if the Rust enum shape changes. The intake fixture is captured with `includes_sensitive_context: true` so both sides emit that field (it is `#[serde(default)]` without `skip_serializing_if`, always emitted Rust-side; TS omits-when-false, which serde `default` accepts — interop-safe, covered separately).

**Minors:**
- **msg_id collision — NON-ISSUE (verified):** each mission call opens a *fresh* sealed session (`runMissionRoundTrip` reconnects per call) and sends exactly *one* envelope. The server's `IdempotencyTracker` is constructed fresh per connection (`serve_sealed_session` stack-local in `hub_agent_run_server`), so it only dedupes *within* a session — it never sees a second envelope on the same session. A stable per-entity `msg_id` therefore cannot be misread as a within-session replay even on repeated transitions. No per-call suffix needed; documented in-code. Cross-handshake replay is defeated by the per-handshake `session_nonce`.
- **truthLabel:** the `*ResultWire` structs carry no `truth_label` field, so the server cannot send one — the hardcoded `rust_wired` is a TS-side ceiling label asserted by construction (not a wire read). Documented with a justified comment rather than asserting a label the server didn't send.

**Untouched (as required):** the `AgentRunResult` / `dispatchRun` / `buildDispatchEnvelope` / resume path — those variants have inline named fields (flat is correct there). Verified zero changed lines touch that path. Owner-gating + flag-OFF dark-safety preserved (wire-shape-only fix).

**Local gates (all green):** `npm run lint` 0 errors, `npm run typecheck` 0 type errors, mission-spine vitest (16 files / 185 tests incl. 6 new golden cross-checks) + routes suite + `npm run test:contracts` (routes 14 / types 12) all pass; the HTTP route-contract snapshot is **unchanged** (the fix does not move it). Branch is born-current with `origin/main`.

PR #861, opened 2026-06-18
Add D20 prepared batch CLI

thesongzhu:
## Summary
- add an offline `prepare-batch` operator CLI path that turns action specs into signable D20 pending-batch JSON
- compute exact action digests through the real Rust gate classification and canonical action bytes
- fail closed for empty batches, malformed fields, duplicate digests, and Irreversible actions such as `run_command`

## Truth label
D20 DARK/offline mechanism only. No operator private key is read by `prepare-batch`; it signs nothing, executes nothing, flips no live flags, and does not prove TRUE operator-in-the-loop. Friday still has no generic action rollback engine; reversible remains a git/worktree-contained space, not undo.

## Tests
- `cargo test -p friday-operator-cli prepare_batch -- --nocapture`
- `cargo test -p friday-operator-cli -- --nocapture`
- `cargo clippy -p friday-operator-cli --all-targets -- -D warnings`
- `git diff --check`

### DIFF
--- a/rust-core/Cargo.lock
+++ b/rust-core/Cargo.lock
@@ -720,6 +720,7 @@ version = "0.0.1"
 dependencies = [
  "friday-core",
  "friday-crypto",
+ "friday-hub",
  "friday-storage",
  "serde",
  "serde_json",
--- a/rust-core/crates/friday-operator-cli/Cargo.toml
+++ b/rust-core/crates/friday-operator-cli/Cargo.toml
@@ -25,6 +25,9 @@ thiserror.workspace = true
 # Wipe transient secret buffers (decoded seed / hex) after use.
 zeroize.workspace = true
 
+[dev-dependencies]
+friday-hub.workspace = true
+
 [[bin]]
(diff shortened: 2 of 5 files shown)


thesongzhu:
Follow-up hardening after local security review:

- `prepare-batch` no longer accepts `mutating` / `base_risk` from JSON. Classification comes from an internal default-Hub built-in registry mirror and still flows through `friday_core::gate::classify_with_reversibility`. Unknown/non-default actions fail closed.
- Parameter binding now uses the Hub canonical parameter encoding (sorted, length-prefixed key/value pairs), not order-preserving JSON.
- Added compiled-CLI coverage for same-logical-params/different-order duplicate rejection.
- Added a dev-only Hub equality test proving prepared `write_file` digest matches `friday_hub::build_request` digest for the default live request shape.

Truth label unchanged: DARK/offline mechanism only; no key read by `prepare-batch`, no signing, no execution, no live flip, no TRUE operator-in-the-loop claim.

PR #1010, opened 2026-06-21
Harden C1/C2 parity capture artifacts

thesongzhu:
## Summary
- bind C1/C2 parity capture artifacts to each flow's expected harness
- require explicit opt-in before custom flow commands can mint artifacts
- label custom-command evidence separately from default harness execution

## Truth labels
- built-DARK capture hardening only; live parity is not claimed
- organic=0 remains true
- no operator key or production DB writes are introduced

## Verification
- npx vitest run --project default test/unit/scripts/ops/friday-c1-c2-tier1-parity-capture.test.ts
- authorized custom artifact smoke check for harness/commandSource/evidence kind
- git diff --check


thesongzhu:
Superseded by #1013, which batches this C1/C2 parity capture hardening with adjacent proof-hardening slices to avoid serial strict-base CI churn. Branch retained for audit.

PR #1011, opened 2026-06-21
Derive health isolation truth from shared status

thesongzhu:
## Summary
- make standalone health capabilities derive execution-isolation truth from the shared status provider
- add an injectable status seam so tests can prove both no-sandbox and partial-sandbox reports without host-dependent assertions
- keep agent.exec and dynamic-import surfaces explicitly non-OS-sandboxed

## Truth labels
- status/reporting closure only; this does not claim full OS sandbox coverage
- skill.shell/python may report partial_os_sandbox only when the shared probe sees sandbox-exec available
- agent.exec, plugin entrypoints, and Node skills remain non-OS-sandboxed in the reported truth

## Verification
- npx vitest run --project default test/unit/skills/executor/friday-execution-isolation-status.test.ts test/unit/skills/executor/friday-shell-executor.test.ts test/unit/api/http/routes/friday-health-routes.test.ts
- npm run typecheck:src
- git diff --check

### DIFF
--- a/src/api/http/routes/friday-health-routes.ts
+++ b/src/api/http/routes/friday-health-routes.ts
@@ -7,7 +7,10 @@
 
 import type { FridayRouteDefinition } from "../../model/friday-api-common.types.js";
 import type { FridayRuntimeCapabilityMatrix } from "#providers";
-import type { FridayExecutionIsolationStatus } from "../../../skills/executor/friday-execution-isolation-status.js";
+import {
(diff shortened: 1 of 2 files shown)


thesongzhu:
Superseded by #1013, which batches this execution-isolation health truth fix with adjacent proof-hardening slices to avoid serial strict-base CI churn. Branch retained for audit.

PR #1012, opened 2026-06-21
Tighten mission spine UI proof readiness gate

thesongzhu:
## Summary
- reject pending real-capture/runtime projection markers in the mission-spine UI/device proof gate
- reject proof artifacts that label themselves organic, preserving the works-vs-adopted truth boundary
- extend the shell self-test for organic-claim and pending-capture rejection

Truth label: mechanism proof readiness only; this does not claim OG9 organic, GO-LIVE, or v1 done.

## Verification
- bash -n rust-core/scripts/mission-spine-ui-device-proof-gate.sh rust-core/scripts/mission-spine-ui-device-proof-gate-self-test.sh
- bash rust-core/scripts/mission-spine-ui-device-proof-gate-self-test.sh
- git diff --check


thesongzhu:
Superseded by #1013, which batches this mission-spine UI proof readiness gate with adjacent proof-hardening slices to avoid serial strict-base CI churn. Branch retained for audit.

PR #1580, opened 2026-07-13
RUN-RESOURCE-CONFLICT: canonicalize tool-batch conflict keys (./, .., trailing-slash, darwin case)

thesongzhu:
## Defect (RUN-RESOURCE-CONFLICT-001 — live)

The agent tool-batch conflict graph decides which `tool_use` blocks may run in parallel. It keyed conflict detection on the **RAW path string**, so two same-turn **MUTATING** tools targeting the **SAME** underlying file via different spellings were wrongly judged non-conflicting → placed in one parallel group → run concurrently via `Promise.allSettled` → **lost-update / interleaved-write corruption** of a real user's file.

`src/agent/runtime/friday-agent-tool-batch-executor.ts`:
- `classifyToolBatchDependencies` conflict-detected via `groupPaths.get(p)` / `groupPaths.set(p, …)` keyed on the raw string, so `/tmp/a.txt` vs `/tmp/./a.txt` (also trailing slash, `..` segments, and macOS case variants) → distinct keys → no conflict → same parallel group.

## Live seam (unconditional)

`friday-agent-runtime.ts` `classifyToolBatchDependencies(executableBlocks)` (inside `if (executableToolUses.length>0 && !aborted)` — no flag) → `executeToolBatch(groups, …)` runs same-group tools in parallel. Consumed by `createFridayAgentRuntime` → the live API runtime.

## Fix (minimal, surgical, no-degrade)

Added `canonicalizeConflictKey(p: string): string` and applied it **only at the conflict-KEY sites** (`groupPaths.get(...)` and `groupPaths.set(...)`):
- `normalize(p)` — folds `.`, `..`, duplicate separators.
- strip a trailing separator (except a bare root `/`).
- lowercase the key **only when `process.platform === "darwin"`** (case-insensitive FS). On Linux the FS is case-sensitive, so distinct-case paths stay distinct.

The literal path passed to the tool executor is **never** rewritten. `extractFilePaths`' public contract is unchanged — its literal output is relied on by other runtime callers (`friday-agent-runtime.ts:4921`, `:7847`) for real file operations (`fileVersionTracker.checkBeforeWrite`), so it was intentionally left untouched.

**No-degrade rationale:** canonicalization can only make MORE paths compare equal → only ADDS serialization, never removes a real conflict; genuinely-distinct files are never merged into incorrect parallelism, and a false-merge merely serializes two writes (both still happen, ordered) — no correctness loss.

## Red-first evidence

Reverting **only** the two `canonicalizeConflictKey(...)` key-site calls (back to raw `p`) makes the normalize assertions throw real `AssertionError`s:

```
× serializes two mutating writes on the same file spelled with a `.` segment
  AssertionError: expected [ { tools: [ …(2) ] } ] to have a length of 2 but got 1
× serializes two mutating writes on the same file with vs without a trailing slash
  AssertionError: expected [ { tools: [ …(2) ] } ] to have a length of 2 but got 1
× serializes two mutating writes on the same file spelled with a `..` segment
  AssertionError: expected [ { tools: [ …(2) ] } ] to have a length of 2 but got 1
× serializes two mutating writes that differ only by case on darwin
  AssertionError: expected [ { tools: [ …(2) ] } ] to have a length of 2 but got 1
```

Restoring the fix → all green (the no-degrade tests — distinct files stay parallel, non-mutating reads on the same canonical path stay grouped — remain green in both states).

## Gate results

- **RED-FIRST PROVEN**: yes (4 real AssertionErrors on revert; green on restore).
- **Full suite green**: executor file (10 passed, 1 skipped) + all `test/unit/agent/runtime/` (606 passed, 1 skipped).
- `npx tsc --noEmit`: **exit 0**.
- `npm run lint`: **0 errors** (pre-existing warnings only; none in changed files).
- `detect-secrets-hook --baseline .secrets.baseline <both files>`: **exit 0** (baseline/excludes untouched).
- **No migration.** Changed files = ONLY `friday-agent-tool-batch-executor.ts` + its test (`git diff --stat origin/main` = 2 files, 121 insertions, 3 deletions). Did NOT touch `friday-agent-runtime.ts` / `friday-agent-run-presentation.ts` / engine adapters (born-current-mergeable in parallel).

## Platform nuance

The darwin case-fold test is guarded with `it.skipIf(process.platform !== "darwin")` and a Linux-only mirror asserts case-distinct paths stay parallel. The platform-agnostic red anchors are the `.`/`..`/trailing-slash cases (pass on Linux CI).

## Scope honesty

Fixes the live tool-batch conflict-key defect. Closes **NO** END-BAR requirement; product stays **NO_GO**.

Born off origin/main `4bbeb48f`.

### DIFF
--- a/src/agent/runtime/friday-agent-tool-batch-executor.ts
+++ b/src/agent/runtime/friday-agent-tool-batch-executor.ts
@@ -14,6 +14,8 @@
  * the default. The batch executor is an opt-in enhancement.
  */
 
+import { normalize } from "node:path";
+
 import { isMutatingToolCall } from "./friday-agent-tool-mutation.js";
 
 // ─── Types ───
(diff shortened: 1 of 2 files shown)


thesongzhu:
Superseded by §12.1 aggregate PR #1582 (same run/execution-loop trust boundary; this branch's commit is absorbed there). Not merged individually per contract §12.1 大而内聚.

PR #1606, opened 2026-07-14
feat(observability): report-only bounded realtime_events growth collector (no route exposure; readback deferred to R3)

thesongzhu:
## Head & review (bound to exact SHA `06e4cb4547eb70f21e8c1ffb04802bfaeabe65d1`)

- **Head SHA:** `06e4cb4547eb70f21e8c1ffb04802bfaeabe65d1`
- **Base:** `main` @ `ef14e5986bb1156c75d49da3f72a97507d337cee`; this PR is **born-current** onto it.
- **Fresh independent exact-SHA review: PENDING** (will be filled by the independent reviewer on this head SHA).
- This head **narrows the PR to a report-only bounded growth COLLECTOR with NO route/HTTP exposure** (see next section). Any earlier review conclusions were bound to superseded SHAs and do **not** carry as a PASS for this head; re-review is required on `06e4cb45`.

---

## What this PR is (after the split): a report-only growth COLLECTOR — NO route/HTTP exposure

Report-only growth observability for the unbounded, append-only, redacted-at-write, **DERIVED** `realtime_events` projection/replay stream (NOT canonical — run recovery / receipts / audit live in other tables). The Hub's `system-health-monitor` computes a **bounded** growth reading every 5 minutes and surfaces it **only via transition-only warning logs**. The signal is **not exposed on any `/v1/observability/*` route or any HTTP surface**.

### Why the split (two P0s avoided)

An earlier revision published the growth reading (row-count / estimated-bytes / status / sample-size / reclaim-status + time-series) onto `GET /v1/observability/metrics` and `GET /v1/observability/time-series`. Two independent Advisor audits found:

1. Both routes were `auth:{public:true}` and absent from the sensitive-read floor, so an **unauthenticated** HTTP GET returned 200 + the growth body.
2. The sensitive-read floor only rejects the **anonymous** principal — it does **not** bind the authenticated principal to the **owner** of this hub-global data, so any authenticated second-user / tenant / device / channel principal would still read owner/system growth data.

Rather than build the owner-authorization model piecemeal now, the operator chose to **SPLIT #1606**: keep the collector, drop the exposure, and **defer the owner-authorized readback to R3** (Settings/API + UI), where the owner / tenant / device / session authorization model is designed holistically.

### What was REMOVED (the exposure)

- `src/observability/services/friday-observability-api-service.ts` — reverted to `main`: no `recordRealtimeEventsGrowth`, no growth gauges, no growth in `getSnapshot()` / time-series.
- `src/hub/friday-hub-bootstrap.ts` — the `metricsSink → observabilityService.recordRealtimeEventsGrowth` publish wiring is removed. The monitor stays **scheduled** and report-only-**logged** (transition-only deduper retained).
- `src/learning/services/friday-system-health-monitor.ts` — the `FridayRealtimeEventsGrowthReporter` / `metricsSink` reporter seam removed; the growth reading now rides only in the run-summary `detail` consumed by the log path.
- `src/api/http/friday-sensitive-read-routes.ts` and its HTTP gate test — reverted to `main`. With no exposure there is no route to gate; re-gating the pre-existing generic `/v1/observability/metrics` + `/time-series` would be an out-of-scope degrade of anonymous diagnostics.

### What was KEPT (the collector value)

- The **bounded** growth computation: O(1) `MAX(rowid)` row-count proxy + `LIMIT`-1000 byte sample; the **correct** `sampleSize` (the real `COUNT(*)` of the sampled subquery, not `min(1000, MAX(rowid))`); `status`; `reclaim_status`.
- **Fail-closed to `degraded`** on any query error; **never** deletes / prunes / vacuums.
- The `DashboardDataProvider` bounded ring buffer (`FRIDAY_MAX_TIMESERIES_POINTS_PER_METRIC = 2000`) — **kept as a general in-memory-retention safety bound** (its test records a generic metric and asserts eviction), independent of the growth feature.
- Retention-job independence: `src/jobs/retention/*` untouched; the DATA-RETENTION-001 negative ("default policy deletes 0 realtime_events rows even after long time-travel") is a **standalone** test.

## Proof (inversion — growth is computed but reaches NO route)

- **Integration** (`test/integration/hub/realtime-events-growth-readback.integration.test.ts`): real `createFridayHub` → `hub.start()` runs the real registered `system-health-monitor` job; the collector is shown to compute a real growth reading from seeded rows; then the **real hub-registered observability routes are served through a real `createFridayHttpServer`** and fetched over HTTP — `/v1/observability/metrics` and `/v1/observability/time-series` carry **none** of the growth fields (`realtimeEventsGrowth`, `friday.realtime_events.*` gauges, `reclaim_status`, growth values). This is a **regression guard**: re-adding the bootstrap publish would fail it.
- **Service seam** (`test/unit/observability/services/realtime-events-growth-readback.test.ts`): a real monitor run + real `createFridayObservabilityRoutes` over a real service — the route handlers carry no growth field; plus a collector **fail-closed + zero-deletion** proof (injected read failure → `degraded`, and every existing row is left **byte-identical** by SHA-256 digest).
- **Collector level** (`test/unit/learning/services/friday-system-health-monitor.test.ts`): correct `sampleSize` under rowid gaps, heuristic classification, true UTF-8 byte semantics, bounded/index-backed queries, bounded budget on 50k+ rows, fail-closed degraded, and transition-only log rate-limiting.

## What this does NOT do (honest disclosures)

- **No queryable readback surface here.** Growth is computed per-tick and surfaced **only** via the monitor's transition-only warning logs. There is no route, API, or durable store for it in this PR.
- **Owner-authorized readback is DEFERRED to R3** (Settings/API + UI), where the owner / tenant / device / session authorization model is designed holistically. This PR intentionally ships **no** growth readback endpoint.
- **Does not expose growth on any route**, and **does not close DATA-RETENTION-001** (still no user control to delete/export `realtime_events`; safe space reclaim remains `deferred_to_rust_epoch_resync`, performed by the future Rust realtime epoch-resync path, not here).

https://claude.ai/code/session_01JhCDTLngxAf6MDh4Bfcd48

### DIFF
--- a/src/api/http/friday-sensitive-read-routes.ts
+++ b/src/api/http/friday-sensitive-read-routes.ts
@@ -32,6 +32,15 @@
  *   - /v1/grants     active grants and authorization posture
  *   - /v1/audit      runtime audit logs / forensic evidence
  *   - /v1/observability/audit  observability audit entries and detail readback
+ *   - /v1/observability/metrics, /v1/observability/time-series  report-only realtime_events
+ *                    DB-footprint growth posture (#1606) — row-count / estimated-bytes / status /
+ *                    sample-size / reclaim-status and its time-series. That is owner/system detail,
+ *                    NOT minimal anonymous health, so it is gated 2026-07-15 (SEC-NET-PRINCIPAL-001).
+ *                    #1606 shipped both routes as auth:{public:true} and omitted them here, so an
+ *                    unauthenticated HTTP GET returned 200 + the growth body. Sibling anonymous
+ *                    observability reads (/v1/observability/overview, /v1/observability/traces,
+ *                    /v1/observability/slos, /v1/observability/alerts) stay public via the
+ *                    trailing-slash boundary (no over-flooring).
  *   - /v1/providers/usage  provider spend/usage summary (global cost data) — gated
  *                    2026-06-24 (operator-authorized per-handler review). Sub-path only:
  *                    the trailing-slash boundary keeps the bare /v1/providers list/get/health/
@@ -82,6 +91,17 @@ export const FRIDAY_SENSITIVE_READ_ROUTE_PREFIXES: readonly string[] = [
   "/v1/grants",
   "/v1/audit",
   "/v1/observability/audit",
+  // SEC-NET-PRINCIPAL-001 (2026-07-15): the report-only realtime_events DB-footprint growth
+  // posture (#1606) is published onto these two routes — row-count / estimated-bytes / status /
+  // sample-size / reclaim-status and its time-series. That is owner/system detail, NOT minimal
+  // anonymous health. #1606 declared both auth:{public:true} and omitted them here, so an
+  // unauthenticated HTTP GET returned 200 + the growth body (bypassing the sensitive-read floor).
+  // Gate the whole route behind a bound principal. The trailing-slash boundary in
+  // isFridaySensitiveReadRoute keeps the sibling anonymous observability reads
+  // (/v1/observability/overview, /v1/observability/traces, /v1/observability/slos,
+  // /v1/observability/alerts, …) public — these two exact paths do not over-floor.
+  "/v1/observability/metrics",
+  "/v1/observability/time-series",
   // Gated 2026-06-24 (operator-authorized per-handler review). Each is a SUB-PATH of an
   // otherwise-anonymous prefix; the trailing-slash boundary in isFridaySensitiveReadRoute keeps
   // the bare /v1/providers and /v1/system/... prefixes anonymous (no over-flooring).
--- a/src/api/http/friday-sensitive-read-routes.ts
+++ b/src/api/http/friday-sensitive-read-routes.ts
@@ -32,15 +32,6 @@
  *   - /v1/grants     active grants and authorization posture
  *   - /v1/audit      runtime audit logs / forensic evidence
  *   - /v1/observability/audit  observability audit entries and detail readback
- *   - /v1/observability/metrics, /v1/observability/time-series  report-only realtime_events
- *                    DB-footprint growth posture (#1606) — row-count / estimated-bytes / status /
- *                    sample-size / reclaim-status and its time-series. That is owner/system detail,
- *                    NOT minimal anonymous health, so it is gated 2026-07-15 (SEC-NET-PRINCIPAL-001).
- *                    #1606 shipped both routes as auth:{public:true} and omitted them here, so an
- *                    unauthenticated HTTP GET returned 200 + the growth body. Sibling anonymous
- *                    observability reads (/v1/observability/overview, /v1/observability/traces,
- *                    /v1/observability/slos, /v1/observability/alerts) stay public via the
- *                    trailing-slash boundary (no over-flooring).
  *   - /v1/providers/usage  provider spend/usage summary (global cost data) — gated
  *                    2026-06-24 (operator-authorized per-handler review). Sub-path only:
  *                    the trailing-slash boundary keeps the bare /v1/providers list/get/health/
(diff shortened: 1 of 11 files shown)


thesongzhu:
**Status: NEEDS_CHANGES — do NOT merge head `daacdbb0`.** External Advisor rejected this SHA. Blocking reasons (all confirmed): FridayMetricsCollector is in-memory (not durable); /v1/observability/metrics enumerates only COUNTER_METRICS so the new gauges are absent from the API; direct setGauge bypasses metricState + dashboard.recordDataPoint; /v1/observability/time-series has none of it and the alert provider can't read it; tests read collector internals rather than bootstrap→production HTTP; status-in-label leaves stale gauge variants — so the durable/readback/time-series/trend claims are not yet true.

A superseding revision is in flight that will (1) write through the formal production observability seam, (2) make rows/bytes/status/reclaim_status authoritatively readable from the real production API, (3) either prove restart-durable readback or honestly relabel as restart-volatile snapshot + trend PENDING, (4) avoid stale label variants, while keeping the verified bounded MAX(rowid) + 1000-row BLOB byte sampling + zero-deletion. It will land as a NEW head SHA with full CI + a fresh independent reviewer (this SHA's prior review/CI are void). Please wait for that SHA.

thesongzhu:
## Readback fix pushed — new head `6be72040`

Re-wired the report-only realtime_events growth signal through the **formal production observability seam**. Addressing each of the 7 rejection points:

1. **In-memory collector, not durable** — Acknowledged and made HONEST. Every "durable" claim removed. The signal is now labeled **RESTART-VOLATILE** (`durability: "restart_volatile"` on the metrics-route snapshot); a durable cross-restart trend is explicitly **PENDING**. This matches the platform's existing model — no metric in this in-memory collector survives restart.
2. **Gauges absent from `/v1/observability/metrics`** — The metrics route now enumerates `GAUGE_METRICS` (rows/bytes/status-code) alongside `COUNTER_METRICS`, and returns a structured `realtimeEventsGrowth` object carrying the status + reclaim_status **strings** (numeric gauges can't). All four fields (rows, bytes, status, reclaim_status) are now authoritatively readable off the real route.
3. **`metrics.setGauge` bypassed metricState + dashboard.recordDataPoint** — Fixed. New `reportGauge()` helper writes through the FULL seam counters/histograms use: `metrics.setGauge` + `metricState.lastValues`/`lastReportedAt` (alert-provider readable) + `dashboard.recordDataPoint` (time-series). The health monitor no longer touches the raw collector; it hands its reading to `observabilityService.recordRealtimeEventsGrowth(detail)`.
4. **`/time-series` had none of this; alert provider couldn't read it** — Now writes to `dashboard.recordDataPoint`, so `GET /v1/observability/time-series` returns the series and `metricState` feeds the alert engine. A multi-tick test advances the clock and asserts the trend `[10, 25, 40]` comes back from the real time-series route.
5. **Tests only `new`'d a collector / never hit the production route** — New suite `test/unit/observability/services/realtime-events-growth-readback.test.ts` builds the service, wires the monitor exactly as bootstrap does, and invokes the **real** `createFridayObservabilityRoutes(...)` handlers for `/v1/observability/metrics` and `/v1/observability/time-series`. The old misleading "durable readback" collector-internals test was replaced with a focused "monitor forwards the reading to the reporter" unit test.
6. **status in the LABEL left stale variants** — Fixed. Status is now a stable **UNLABELED** gauge VALUE (`growth_status_code`: 0 healthy / 1 warn / 2 critical / 3 degraded). A transition test drives healthy→warn→critical→healthy and asserts `getAllSnapshots(gauge)` has exactly **one** entry (no leaked variants) and the route reflects only the CURRENT status.
7. **"durable/readback/time-series/trend" wording was untrue** — Corrected throughout. Now: within-session trend is LIVE and proven via the route; durable cross-restart trend is labeled PENDING; snapshot labeled restart-volatile.

### Durable vs volatile (explicit)
**RESTART-VOLATILE.** The observability metrics collector and its time-series store are in-memory with no persistence path (verified: no metrics/time-series table, no shutdown flush). A restart test proves a fresh service instance returns `realtimeEventsGrowth: null` and zeroed gauges. Durable trend is PENDING.

### Red-first evidence
A production-path probe (daacdbb0 APIs only) asserting `GET /metrics` returns `friday.realtime_events.rows_estimate` **FAILED before** the fix: `AssertionError: expected undefined to be 15`. After the fix the full readback suite is GREEN (6/6).

### Preserved (unchanged)
Bounded `MAX(rowid)` row proxy (no full-table COUNT), 1000-row `LENGTH(CAST(payload_json AS BLOB))` byte sampling (no full-table SUM), transition-only log dedup, and **ZERO deletion** (independent retention negative test still green; degraded path is fail-closed and delete-free).

### Tests / gates
`tsc --noEmit` PASS · ESLint 0 errors on touched files · 341 tests green across observability + health-monitor + observability-routes + retention (incl. the 6 new production-path readback tests).

### does_not_prove
- Does NOT prove durability — the signal is restart-volatile by design; a durable, persisted growth trend is out of scope (PENDING, deferred to a future Rust-owned metrics-persistence epoch).
- Does NOT delete or reclaim any realtime_events rows — reclaim remains `deferred_to_rust_epoch_resync`.
- Route handlers are invoked directly (the exact functions the HTTP server binds); a full network round-trip through the live Hub was not exercised.

PR #1607, opened 2026-07-14
fix(memory-guard): redact full-width PII on the memory egress path incl. appended learned facts

thesongzhu:
## Requirement
No PII on the memory read/egress path (`GET /v1/memory/items[/:id]`, `POST /v1/memory/search`): credit cards, phones, SSNs must be redacted in the returned HTTP JSON across **content, metadata, tags, AND snippet** — including **full-width** forms and including **appended learned facts** — with **no regression** (never redact fewer chars than before).

## Frozen head SHA
`c851a6350aaeaf39cddf7dbaa6adca1cbe0c11b5` (base `d7bb8dc7513f9e33131f0947d754cb7217e02748` = current origin/main). Supersedes `496e0d27`/`a2049bd2`, whose evidence is void.

## LIVE production seam (egress / read path)
`GET /v1/memory/items/:id`, `GET /v1/memory/items` (list), `POST /v1/memory/search` (`friday-memory-routes.ts`) → the guard service output filter (`filterItem`/`filterSearchResults` → `piiRedactor`). These READ paths are **not** gated by `tsMemoryWritesEnabled`, so leaks here are live.

## Two leaks fixed
**Leak A — full-width forms bypass the ASCII-only regexes.** `\d = [0-9]` (no `u` flag) and `luhnCheck` strips to `[0-9]`, so a Luhn-valid card in full-width digits (U+FF10–FF19) was returned verbatim.

**Leak B (HIGH, found by review) — learned facts appended after the filter.** `memory.list` (`:594`) and `memory.search` (`:534`) append `deps.listLearnedFacts(...)` to the result **after** the guard's output filter has already run, so learned-fact `content` / `metadata` / `tags` / `snippet` skipped PII redaction entirely (any PII, full-width or ASCII, leaked in the JSON).

## Fixes
[...]

PR #1610, opened 2026-07-15
fix(memory): redact learned-fact PII on sibling egress routes (uix/asset-inventory/agent-tool)

thesongzhu:
## req #61 LEARNED-FACT-EGRESS-FILTER (+ PII)

**Head SHA:** e67a91a5936ea0d043098f27dbc39a59c3606086
**Base:** main @ ddc379d6 (#1607, merged)
**Reviewer:** Advisor PASS @ `e67a91a5936ea0d043098f27dbc39a59c3606086` — code audit PASS, required CI green, independent focused tests 16/16 PASS; scope confirmed (3 sibling value leaks only).

### What & why
PR #1607 added a route-level PII output filter to the memory read routes (`GET /v1/memory/items`, `POST /v1/memory/search`) so appended **learned facts** — written verbatim, bypassing the write-time guard — are redacted at egress. The **same raw learned-fact value** still egressed **UNREDACTED** through three sibling paths that read `deps.listLearnedFacts` directly:

1. `GET`/`PATCH` `/v1/uix/learned-facts` — `src/api/http/routes/friday-uix-routes.ts` (~536, ~581)
2. `GET /v1/assets/inventory` — `src/api/http/routes/friday-asset-inventory-routes.ts` (~112, `details.value`)
3. `memory_search` **agent TOOL** — `src/agent/tools/friday-agent-memory-tools.ts` (~759-769) — a **trust boundary** to the agent

### Fix — reuse the #1607 production filter (no reimplementation)
- **New** `FridayMemoryGuardOutputFilter.redactLearnedFactValue(value)` in `friday-memory-output-filter.ts`, delegating to the SAME production deep redactor (`redactDeep`) that #1607's `redactMetadata` uses. It preserves the free-form `value`'s structure/type (string, nested object/array) and redacts PII in place. **No stringify/truncate** — these sibling routes return the RAW `value` field (avoids degrade). Idempotent.
- **uix** (list + update): redact learned-fact `value` before `enrichLearnedFactBoundary` (which already strips metadata → `value` is the only carrier).
- **asset-inventory**: redact `details.value` on the `learned_fact` item.
- **agent tool**: route the learned-fact search result through the SAME `filterSearchResult` (#1607) as the final egress transform.

### Agent-tool trust-boundary decision → **DEFAULT = FILTER (redact)**
No hard functional reason for the agent to receive raw PII (email/phone/ssn/card) from a learned *preference* fact — that raw value is exactly what must not cross the boundary. Applied to **learned facts only**; stored results come from `memoryService` and are **unchanged** (out of scope; changing stored egress would broaden scope and risk degrading legitimate agent memory access). **No remaining leaf registered** — the learned-fact value is filtered.

### red-first evidence (production route handler + agent tool oracle, not mocks)
With the egress wiring reverted (new filter method present but unused), the 3 path specs fail with the **raw** value egressing verbatim; re-applying the wiring makes them pass:
- uix list/update — before: `カード番号は４１１１１１１１１１１１１１１１です` (raw full-width card) / `alice@example.com`; after: `カード番号は[CREDIT_CARD]です` / `[EMAIL]`
- asset `details.value` — before: raw full-width card / raw email; after: `[CREDIT_CARD]` / `[EMAIL]`
- agent `memory_search` content — before: raw full-width card; after: `[CREDIT_CARD]`

Also covered: nested metadata/arrays (deep), idempotency (filter-once stable), Unicode full-width width-fold, a perf-sanity bound (no O(n^2)), and negative controls (no-PII values pass unchanged; unrelated stored agent result unchanged).

### Tests / checks
- 4 new specs, 16 tests — green. Existing regression suites for touched files (uix, agent tools, asset-inventory, memory output-filter, memory-pii-guard, #1607 route test) — 141 tests green.
- `tsc --noEmit` (src) — 0 errors. `eslint` — 0 errors on touched files (pre-existing warnings only).
- `src/hub/friday-hub-bootstrap.ts` **NOT** modified (PR #1606 owns it).

### does_not_prove
This closes the **3 sibling learned-fact egress leaves only**. Overall PII requirement is **NOT** closed: the realtime payload redactor, other Unicode scripts, write-scan-dark, key/id fields, and deep-object leaves remain **OPEN** (numeric-typed value + object-key coverage tracked as follow-up). No expansion of closure claim.

### DIFF
--- a/src/agent/tools/friday-agent-memory-tools.ts
+++ b/src/agent/tools/friday-agent-memory-tools.ts
@@ -5,6 +5,7 @@ import type {
   FridayMemoryService,
 } from "#memory";
 import type { FridayMemoryItem, FridayMemorySearchResult } from "#memory";
+import { createFridayMemoryOutputFilter } from "#memory";
 import type { FridayLearningEventAppendInput } from "#ledger";
 import {
   errorResult,
@@ -52,6 +53,14 @@ export interface CreateFridayAgentMemoryToolsDeps {
 }
 
 // ─── Namespace scoping ───
+
+// Learned facts are appended to memory_search results AFTER being written verbatim (they
+// bypass the write-time PII guard). memory_search egresses across a TRUST BOUNDARY to the
+// agent, so the learned-fact result is routed through the SAME production PII output filter
+// (#1607) before it reaches the tool caller — redacting content/metadata/tags/snippet. This
+// applies ONLY to learned facts; stored results come from `memoryService` unchanged (out of
+// scope for this fix).
+const memoryLearnedFactOutputFilter = createFridayMemoryOutputFilter();
 
 const AGENT_MEMORY_BASE_NAMESPACE = "agent";
 const USER_FACING_MEMORY_NAMESPACES = new Set(["default", "user", "preference"]);
@@ -759,7 +768,8 @@ function createMemorySearchTool(
         const learnedResults = principalId && deps.listLearnedFacts && shouldIncludeLearnedFacts(explicitNamespace)
           ? deps.listLearnedFacts({ userId: principalId, limit })
             .filter((fact) => matchesLearnedFactQuery(fact, query))
-            .map((fact) => toLearnedFactSearchResult(fact, query))
+            // Egress PII filter across the agent trust boundary (learned facts only).
+            .map((fact) => memoryLearnedFactOutputFilter.filterSearchResult(toLearnedFactSearchResult(fact, query)))
           : [];
         const combined = [...dedupedResults, ...learnedResults]
           .sort((a, b) => b.score - a.score)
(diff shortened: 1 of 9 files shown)


thesongzhu:
**Independent review: PASS** @ e67a91a5 (fresh reviewer, role-separated from implementer). Traced all 3 sibling paths — the free-form learned-fact `value`/`content` is routed through the same #1607 production redactor at the real egress point; nested string coverage correct; no missed sibling field; agent-tool default-redact boundary sound; no regression; scoping honest.

**Registering the reviewer's Low residual leaves explicitly (dispositioned, not silently omitted):**
- **Numeric-typed values** — `redactDeep` scans only string leaves, so a bare-number learned-fact value (e.g. `value: 4111111111111111`) would egress raw on the two *structured* routes (uix list, asset-inventory). The agent-tool path is immune (JSON.stringify → string → scanned). This is the **same pre-existing `redactDeep` limitation shared with #1607's `redactMetadata`** (merged), and requires the learning extractor to store a bare number rather than text (atypical). Registered as an OPEN leaf.
- **Object string *keys*** not redacted (only values) — exotic shape; OPEN leaf.
- **Nesting depth > 6** on the structured routes — already disclosed as 'deep-object leaves remain OPEN'.

A considered fix for numeric/typed-PII + key redaction in the shared `redactDeep` (degrade-aware, since it also affects #1607) is filed as a separate follow-up — NOT bolted onto this PR to avoid scope-expansion and false-positive redaction of legitimate numbers. This PR's stated scope (string-shaped full-width/CJK/ASCII PII on the 3 sibling routes) is genuinely closed. Overall PII requirement remains OPEN.

PR #1623, opened 2026-07-19
SEC/DUR: HTTP operation-journal residual hardening — fail-closed on receipt-write failure + never persist replay secrets (DUR-OPERATION-JOURNAL-001)

thesongzhu:
## SEC/DUR — HTTP operation-journal residual hardening (DUR-OPERATION-JOURNAL-001)

Two CONFIRMED HIGH residual defects at the shipped HTTP operation-journal's own write boundary, found by an adversarial residual-gap hunt after independently re-executing the acceptance oracle (cross-store-digest-conflict 4/4, usage-receipt 3/3, operation-journal-durability e2e 2/2 — all green on frozen main). Batched into one seam-scoped PR. **Code-only — no migration, no schema, no MEMORY.md change.**

### Defect #1 [HIGH] — transient `complete()` failure RELEASED the reservation → retry re-executes the committed effect
`friday-http-server.ts`: the handler commits its durable side-effect, then `idempotencyStore.complete()` writes the receipt in a SEPARATE transaction (the TS write path has no busy-retry). If that receipt write threw (`SQLITE_BUSY`/`IOERR` AFTER the effect committed), the generic catch called `release()` (DELETE) → a retry with the same key found no row → re-reserved → **RE-EXECUTED** the already-committed effect. Strictly weaker than the process-crash path (crash leaves `in_flight` → boot `reconcileOrphanedReservations()` → `indeterminate`; active release defeats that). This is exactly the requirement's negative control.

**Fix:** new store method `markIndeterminate(key)` (`UPDATE … SET status='indeterminate' WHERE …=? AND status='in_flight'` — row kept, never deleted, scoped so a completed receipt is never clobbered); a `handlerCommitted` flag set immediately after `await route.handler(ctx)`; in the catch — committed → best-effort `markIndeterminate` (fail-closed; retry refused, never re-executes), falling back to leaving `in_flight` for boot reconcile; threw-pre-commit → `release()` (retryable, unchanged). SSE / raw-text SUCCESS-path releases left intact (response already sent — distinct from the error path).

### Defect #2 [HIGH] — raw secrets persisted in the cached replay body at rest for 24h
`complete(..., data: result)` stored the full handler return verbatim into `http_operation_journal.response_json` (24h TTL). Idempotency-guarded non-GET routes returning secrets (auth.login/refresh raw access+refresh tokens) leaked secret material at rest, breaking the auth service's deliberate hash-only invariant (`auth_access_tokens` stores only `hashToken(refreshToken)`).

**Fix (general, fail-closed, reuses the canonical detector — no per-route enumeration):** at complete() store-time, if `findSecretShapeSpans(JSON.stringify(result))` fires on the flattened result, store a `{__fridayNonReplayable:true}` sentinel instead of the raw body. The response SENT to the first caller is UNCHANGED (real tokens); only the cached-for-replay copy is sanitized. Replay of a sentinel → typed 409 `SECURITY_IDEMPOTENCY_NONREPLAYABLE`. Durability / no-duplicate-effect is fully preserved; only the secret-body replay convenience is withheld (for a bearer token, a security improvement).

**Precise scope (honest — shape/credential-label based, not all-strings), empirically verified:** the confirmed auth leak IS caught — a realistically-minted pair (accessToken = `<base64url(claims)>.<HMAC>` starting `eyJ`, 220 chars; refreshToken = UUID) makes the detector FIRE on the full login response (via both the `eyJ` shape and the `accessToken`/`refreshToken` credential labels; even the bare-UUID refreshToken fires via its label). Benign `id`/`runId`/`sessionId` UUIDs stay clean → low false-positive. Request-side is safe (request body stored only as SHA-256 `payload_digest`, never raw). **Documented residual (pre-existing detector scope, not a regression — strictly reduces at-rest leakage from 100% raw):** an opaque secret under a non-credential generic key (e.g. `{"value":"<opaque>"}`) would not be flagged; no such concrete guarded route exists (secrets.create returns `toSummary()` metadata, not the raw value). Tracked, not introduced here.

### Red-first (both fail pre-fix, pass post-fix — verified by reverting the fix)
`test/integration/idempotency/friday-idempotency-guard-residual-hardening.test.ts` drives the real `FridaySqliteOperationJournalStore` (in-memory better-sqlite3, v100 migration) through the real `createFridayHttpServer` guard.
- #1: reserve→complete where complete() throws a simulated `SQLITE_BUSY` → asserts the row is NOT deleted (status=`indeterminate`), `effectCount===1`, retry → 409 INDETERMINATE, effectCount still 1 (no re-execute). Pre-fix: `release()` deletes the row → retry re-executes.
- #2: a route returning a detector-flagged token + Idempotency-Key → asserts (a) first response has the real token, (b) `response_json` has no raw token (holds the sentinel), (c) replay → 409 NONREPLAYABLE. Pre-fix: raw token in `response_json`.

### Independent verification
Fresh independent reviewer (refute-first, separate context) → **PASS**, no blocking findings. New tests 2/2; no-regression suite (cross-store-digest-conflict + operation-journal-durability e2e + usage-receipt + pre-existing http-server-idempotency) all green; `tsc --noEmit` 0; eslint 0 errors; ssd-lint pass; detect-secrets pragma present, `.secrets.baseline` untouched. Born-current on main `63f873a4`. 3-file diff (2 source + 1 test).

### Deferred same-family follow-ups (documented, NOT folded in — proportionality / different files / not attacker-forceable)
- #4 [MED-LOW] NULL `payload_digest` short-circuits the conflict guard on legacy pre-v100 rows (projector-service + outbox-queue-service — different files; not attacker-forceable, fresh writes always set a digest).
- #5 [LOW] `learning_events` INSERT OR IGNORE no-digest (learning-signal ledger, not a charge/send). #6 [INFO] Rust `provider_session_event` plain INSERT (fail-closed throw, untyped).

### Governance
R13 preserves R12 §34 U19: implementer ≠ fresh reviewer ≠ merge executor. Builder executes the returned protected `--match-head-commit` squash after exact-SHA PASS + unchanged-recheck. Product NO_GO — this closes a residual on a P0 seam; it does not on its own establish product GO, nor bind the requirement-level `final_artifact_proof` into the hash-bound Handoff ledger (that remains an operator / prework-owner action).

### DIFF
--- a/node_modules
+++ b/node_modules
@@ -1,1 +1,0 @@
-/Users/jarvis/Projects/Friday/node_modules
--- a/node_modules
+++ b/node_modules
@@ -1,0 +1,1 @@
+/Users/jarvis/Projects/Friday/node_modules
--- a/src/api/http/friday-http-server.ts
+++ b/src/api/http/friday-http-server.ts
@@ -770,12 +770,12 @@ export function createFridayHttpServer(deps: FridayHttpServerDeps): FridayHttpSe
     const rawMethod = (req.method ?? "GET").toUpperCase();
     const isHead = rawMethod === "HEAD";
 
-    // Declared outside the request try so the catch block can release an in-flight idempotency
-    // reservation if the handler threw BEFORE committing its side-effect, or fail it CLOSED
-    // (indeterminate, never released) if the handler already committed and only the completed-receipt
-    // write failed. `handlerCommitted` records that boundary.
+    // Declared outside the request try so the catch block can fail the idempotency reservation
+    // CLOSED on ANY post-reservation error. Once we hold a reservation and have invoked the handler,
+    // a thrown handler MAY already have committed a durable effect (a rejected promise is NOT proof
+    // that no effect committed), so the reservation transitions only to `completed` (success) or
+    // `indeterminate` (failure) — it is NEVER released/deleted, which would let a retry re-execute.
     let idempotencyStoreKey: string | undefined;
-    let handlerCommitted = false;
 
     try {
       // Handle CORS preflight (before casting to FridayHttpMethod)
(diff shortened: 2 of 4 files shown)


thesongzhu:
### Updated → `5b787228` (Advisor round 2, addresses NEEDS_CHANGES)

The Codex Advisor caught a HIGH defect the first head (`a3e7f012`) missed: my `handlerCommitted`-after-`handler()`-returns flag was a promise-resolution heuristic — a handler that **commits a durable effect and then throws** left `handlerCommitted=false`, so the catch `release()`d the reservation and a retry re-executed the committed effect (only the `complete()`-throw slice was fixed).

**Whole-class fix (§36.8(5)):** removed `handlerCommitted`. Once `reserve()` succeeds and the handler is invoked, the reservation transitions ONLY to `completed` or `indeterminate` — **never released**. All three post-reservation release sites (catch, SSE headers-sent, raw-text) now fail closed via `markIndeterminate`.

**Red-first `#1b`:** handler commits effect then throws → row kept + `indeterminate`, retry → 409, effect not re-executed (verified RED on `a3e7f012`: row deleted). No-regression: 1421 tests across `test/unit/api/http` + idempotency + durability e2e pass; tsc 0, eslint 0. Code-only, born-current on `63f873a4`.

PR #1624, opened 2026-07-19
SEC/ART-NONPROD-001: gate FRIDAY_ALLOW_TEST_ONLY_* switches behind protected-release profile (no activatable mock lane in release)

thesongzhu:
## SEC/ART-NONPROD-001 — gate `FRIDAY_ALLOW_TEST_ONLY_*` switches behind a protected-release profile

**Requirement (ART-NONPROD-001, P0):** no mock/test-only activation switch may be activatable in the default/production (release) path.

**Concrete defect (found by a currentness reconcile of the next ranked P0 rows after DUR-OPERATION-JOURNAL-001 merged):** three ENV-sourced test-only switches were honored WITHOUT a protected-profile gate, so setting them in a production/tagged-release build activated a test/mock lane:
- `FRIDAY_ALLOW_TEST_ONLY_PLUGIN_EXECUTION` / `FRIDAY_ALLOW_TEST_ONLY_AUTONOMY_LIFECYCLE_EXECUTION` — `resolveTestOnlyFlagFromEnv` in `friday-hub-bootstrap.ts` → drove `pluginRuntimeMode="full"` + test-only plugin/autonomy execution.
- `FRIDAY_ALLOW_TEST_ONLY_MASTER_KEY_GENERATION` — `getMasterKey` in `friday-secret-crypto.ts` → the test-only master-key generation escape hatch (default is fail-closed; production must supply `FRIDAY_MASTER_KEY`).

**Fix (fail-closed, consolidated — mirrors the existing `resolveFridayCanonicalMutatingActionGate` throw precedent):** code-only, no migration.
- New leaf module `src/security/friday-protected-profile.ts`: single-source `isFridayProtectedReleaseProfile(env)` (`NODE_ENV=production` OR `FRIDAY_RELEASE_TAG` non-empty), a CLOSED registry of the test-only ENV switch names, and a fail-closed message naming the specific switch. Leaf module — imports nothing from the Friday tree, so both `#hub` and `#providers` consume it without an import cycle.
- `friday-hub-bootstrap.ts`: `resolveTestOnlyFlagFromEnv` now throws when a truthy value (env-sourced, and defense-in-depth an explicit config `true`) resolves in a protected profile; threaded through both call sites; `isFridayCanonicalGateProtectedProfile` delegates to the shared predicate (single source of truth).
- `friday-secret-crypto.ts`: `getMasterKey` refuses (FridayDomainError 503) when the test-only generation flag is set in a protected profile — scoped AFTER the real-`FRIDAY_MASTER_KEY` return, so a production deploy with a real key is unaffected; only the generation escape hatch is refused.

**Open-world guard:** `test/unit/security/friday-protected-profile.test.ts` enumerates every `FRIDAY_ALLOW_TEST_ONLY_*` literal referenced under `src/` and asserts each is registered AND refused in a protected profile — so a future un-gated test-only switch fails the guard rather than silently shipping an activatable mock lane. The real signed-artifact / release-manifest observation scan is release-gated and intentionally OUT OF SCOPE (code-only source-level guard; mirrors the deferred real-observation boundary in `tools/inventory/reconcile.mjs`).

**Red-first:** with the 2 gate files reverted to `51684b98` (keeping the new leaf module), the 8 throw-expecting cases FAIL (`expected [Function] to throw` — the switches activate) across both `NODE_ENV=production` and `FRIDAY_RELEASE_TAG` profiles and all 3 switches; green after the fix. Dev/test-profile cases stay green (dev/test lanes unaffected). Independently re-verified.

**No-regression:** no existing test relied on activating a test-only switch under a production profile (checked hub-bootstrap-integration, secsetup, closure-lib). New tests + existing hub-bootstrap-env + secret-crypto suites pass; broad `test/unit/hub/` + `test/unit/providers/security/` + `test/unit/security/` = 925 passed; integration + closure-lib 52 passed; tsc 0; eslint 0. Born-current on `51684b98`; 6-file diff, no migration.

**Governance:** R13 preserves R12 §34 U19: implementer ≠ fresh reviewer ≠ merge executor. Builder executes the returned protected `--match-head-commit` squash after exact-SHA PASS + unchanged-recheck. Product NO_GO.

### DIFF
--- a/src/hub/friday-hub-bootstrap.ts
+++ b/src/hub/friday-hub-bootstrap.ts
@@ -46,6 +46,10 @@ import type { FridayStateRuntime } from "#state";
 import { createFridayLocalDaemonService } from "#daemon";
 import { createFridayMutatingActionGate } from "../security/friday-mutating-action-gate.js";
 import {
+  fridayProtectedProfileTestOnlySwitchMessage,
+  isFridayProtectedReleaseProfile,
+} from "../security/friday-protected-profile.js";
+import {
   buildFridayRuntimeCapabilityMatrix,
   createFridayProviderCostCalculator,
   createFridayProviderPricingCatalog,
@@ -870,10 +874,14 @@ export function resolveFridayHubConfig(
   const allowTestOnlyPluginExecution = resolveTestOnlyFlagFromEnv(
     input.allowTestOnlyPluginExecution,
     env.FRIDAY_ALLOW_TEST_ONLY_PLUGIN_EXECUTION,
+    "FRIDAY_ALLOW_TEST_ONLY_PLUGIN_EXECUTION",
+    env,
   );
   const allowTestOnlyAutonomyLifecycleExecution = resolveTestOnlyFlagFromEnv(
     input.allowTestOnlyAutonomyLifecycleExecution,
     env.FRIDAY_ALLOW_TEST_ONLY_AUTONOMY_LIFECYCLE_EXECUTION,
+    "FRIDAY_ALLOW_TEST_ONLY_AUTONOMY_LIFECYCLE_EXECUTION",
+    env,
   );
   const pluginRuntimeModeRaw = input.pluginRuntimeMode ?? env.FRIDAY_PLUGIN_RUNTIME_MODE ?? "stub";
   const pluginRuntimeMode = pluginRuntimeModeRaw === "full"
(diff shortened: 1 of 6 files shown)


thesongzhu:
### PARKED (not merging) — ART-NONPROD-001 closure is artifact-bound + native + signed-release-gated

The Codex Advisor exact-SHA audit (`6cdf248c`) correctly found this env-marker gate does **not** close ART-NONPROD-001, and independently reproduced it by building the source-release artifact:
1. **Artifact-bound gap:** the packaged launcher `scripts/ops/friday-service-run.sh` sets `FRIDAY_CANONICAL_GATE` but neither `NODE_ENV=production` nor `FRIDAY_RELEASE_TAG` — so this gate is a **no-op in the real release path** (`FRIDAY_ALLOW_TEST_ONLY_PLUGIN_EXECUTION=1` still activates the test lane). ART-NONPROD-001 requires the protection to be intrinsic to the release artifact, not dependent on mutable caller env.
2. **macOS native seam:** `FridayHubConsoleApp.swift:67` honors `FRIDAY_CONSOLE_MOCK` / `--use-mock-read-client` unconditionally in the release app — an `apps/macos/**` (native-track) + signed-artifact concern.
3. Open-world guard proves registry membership but not per-consumer enforcement.

Findings #1 (artifact-bound release-profile) + #2 (macOS mock-seam in Swift) + the signed-artifact negative-control span **release-packaging, macOS-native, and code-signing** — operator/release/native/signing-gated. Merging this env-marker slice under a "no activatable mock lane in release" claim would over-state (proof-theater), so it is **PARKED** (branch/head/history retained). The gate here is a genuine **defense-in-depth** hardening (it does refuse the switches when a protected-profile marker is present) and is revivable once the artifact-bound release-profile enforcement + the macOS console-app mock-seam are operator-authorized (native/release/signing scope).

